Issue #4, opened 2020-10-13
Implement a recipeModule manager


rishabhpoddar:
Should be responsible for:
- Provide [routing functionality](https://github.com/supertokens/supertokens-auth-react/issues/4#issuecomment-709993403). The algorithm is [here](https://github.com/supertokens/supertokens-auth-react/issues/4#issuecomment-712193998).

- Has a way for any module to get the `appInfo` if needed.

- Has an init function that takes:
   - The [appInfo](https://github.com/supertokens/supertokens-auth-react/issues/4#issuecomment-709977157)
   - `recipeList`: RecipeModule[]

- Provides an abstract `recipeModule` class that modules must `extend`:
   - Provides methods to asks a module if they can handle the current route (excluding `websiteBasePath`).
   - Get and manage the module's `rId`. The `rId` will be an argument to the constructor.
   - Exposes a function that modules must implement to handle the current route.

- Recipe modules must be build such that parent recipe's `rId` can be propagated to them. [This is how](https://github.com/supertokens/supertokens-auth-react/issues/4#issuecomment-709991055)

rishabhpoddar:
@NkxxkN please review the above

rishabhpoddar:
## AppInfo in the init function:
```
{
   appName: string,
   websiteDomain: string,
   apiDomain: string
   logoFullURL?: string, // Default is ""
   apiBasePath?: string, // Default is `/auth`
   websiteBasePath?: string // Default is `/auth`
}
```

rishabhpoddar:
## If a recipe is being built on top of another one, how can its `rId` be propogated down?

### Base recipe
```jsx
class EmailPassword extends RecipeModule {
	
	static instance: undefined | EmailPassoword

	constructor(options: {}, recipeId) {
		super(recipeId)
	}

	static getInstanceIfDefined() => {
		if (instance === undefined) {
			throw error("...");
		}
		return instance;
	}

	static init(options: {..}): ModuleInterface {
		instance = new EmailPassowrd(options, "emailpassword")
	}
	
}

// feature
class SignInUp extends React.Component {

	getEmailPassowordInstance = () => {
		let instance;
		if (this.props.__internal !== undefined && this.props.__internal.instance !== undefined) {
			instance = this.props.__internal.instance;
		} else {
			instance = EmailPassword.getInstanceIfDefined();
		}
		return instance;
	}
	
	defaultCallSignUp = () => {
[...]

PR #11, opened 2020-10-16
Recipe module manager

NkxxkN:
## Recipe Module Manager (In progress)

#4 

- [x] Init
  - [x] appInfo
  - [x] recipeList
- [x] handleRoute
- [x] getComponentForRoute
- [x] `SuperTokensRoute` react-dom-router` component
- [ ] Allowing to extend features (EmailPasswordModified)  (Future PR, not linked to recipe module manager directly)


rishabhpoddar:
- I'm not sure that changing the `tsconfig.json` `target` from `es5` to `es6` is a good idea. I remember having issues with this for the `supertokens-website` repo and at the end, using `es5` worked everywhere. So maybe we should stick to that?
- I have changed `npm run build` to `npm run build-pretty` when we run the `npm run test` command (pushed to your branch).
- The `recipeIdGetParam` in `constants.ts` should be `"rid"`. Sometimes, in headers, capital letters are automatically transformed to all smaller cases by some backend frameworks. So it's safer to be all small when used in header. Hence, we should use small for when it's in get params as well. Also, since this is a constant, I prefer the name be something like `RECIPE_ID_GET_PARAM`. Talking about names, I prefer "get param" be "query params" - but it's fine now as well :)
- The contents of `version.ts` can be in `constants.ts`?
- The values of `apiBasePath` and `websiteBasePath` can be in `constants.ts` as `DEFAULT_API_BASE_PATH` and `DEFAULT_WEBSITE_BASE_PATH`
- With typescript, I generally prefer using types everywhere instead of an interface, unless you intend that some class `implements` it. So `AppInfo`, `SuperTokensConfig`, should be a type, declared in a `types.ts` file.
- I generally don't like the use of implicit `if` conditions, so `if (SuperTokens.appInfo ||` should actually become `if (SuperTokens.appInfo === undefined ||`. Implicits always cause an issue in a large project.
- If using `if` statements, please always surround them with `{}`, even if just one line. Again, no scope for errors.
- `getUrlFromDomain` needs to follow a certain standard everywhere (notice how none of the outputs have a trailing slash):
   ```
   example.com => https://example.com
   example.com/ => https://example.com
   http://example.com => http://example.com
   https://example.com/ => https://example.com`
   localhost ⇒ http://localhost
   IP Address ⇒ http://IP Address
   if there is a path given, remove it
   ```
   You can also rename the function to `normaliseDomain` or to `cleanDomain` (if changing to `normalise`, then also change `cleanPath` please)
- There is a distinction between the input type of `AppInfo` and what the actual `SuperTokens.appInfo` stores. The input type has a few optional params, whilst the stored result has nothing optional. If we have the same type of both, then whenever we use the store one, we have to do undefined checks for those optional params (and they will never be undefined)
- I prefer if we use the singleton pattern for `SuperTokens` class. This way, `appInfo` will never have to be `undefined`, and there are other benefits too.
- The `cleanPath` must follow the following:
   ```
   custom/base => /custom/base
   custom/base/. => /custom/base
   /custom/base => /custom/base
   custom/base/ => /custom/base
   custom/base/ => /custom/base
   custom-base => /custom-base
   If there is a domain, or query params, or anything else, remove it
   ```
- There are many functions in `SuperTokens` that are exposed to the user, which are not required. For example, `prependWebsiteBasePath`, and `reset`.
- I prefer calling the file `module.ts` => `recipeModule.ts`. Since the class is called `RecipeModule`. Also, I think there is an `abstract` keyword in typescript. If yes, that class should be `abstract`.
- Use `string[]` instead of `Array<string>`
- There is no need to have `getInstanceIfDefined` and `init` in the `RecipeModule` class. These are static functions and should only be in the actual module class.
- `RecipeModule` must have a `getRecipeId` function (which does not need to be implemented in the child classes).
- I think the `handleRoute` function does not follow the algorithm for routing as mentioned in [here](https://github.com/supertokens/supertokens-auth-react/issues/4#issue-720769261). Please correct me if I am wrong.
- In `EmailPassword`, the constructor should take a `recipeId` as described [here](https://github.com/supertokens/supertokens-auth-react/issues/4#issuecomment-709991055)
- In prefer not to use `any` type anywhere. An example of where you are using it is in `EmailPassword`: `signInAndUpFeature: any;`. But I think that's cause it's work in progress
- There is a grammatical mistake in an error message: `Make sure to call "init" method.` => `Make sure to call the "init" method.`
- You do not need `recipeId` and `routes` as static variables in `EmailPassword`, as they are already there in the super class as instance variables.

NkxxkN:
- [x] es6 => es5, fixed (I changed after I got an error but it seems that it's not useful anymore)
- [x] lowercase headers. (`rId` => `rid`)
- [x] Use abstract class for recipeModule, rename file.
- [x] constants to uppercase
- [x] Interface => Type
- [x] Use string[] instead of Array<string>
- [x] typo in error message
- [x] surround `if` with {}, even if just one line. 
- [x] getUrlFromDomain needs to follow a certain standard everywhere
- [ ] `any` types, WIP

> There are many functions in SuperTokens that are exposed to the user, which are not required. For example, prependWebsiteBasePath, and reset.

I removed `prependWebsiteBasePath` for now. And reset is useless in production as it requires to be in testing environment. 
How can we reset appInfo without such a function (without actually exposing the `appInfo` attribute) to make sure we do not expose a useless function? 

> I prefer if we use the singleton pattern for SuperTokens class. This way, appInfo will never have to be undefined, and there are other benefits too.

Using singleton here would feel odd on import:
```js
import SuperTokens from 'supertokens-auth-react';

const supertokens = SuperTokens.init({...})
supertokens.handleRoute();
// OR
SuperTokens.init({...})
SuperTokens.getInstance().handleRoute();
```

OR  in `./index.ts`
```js
class SuperTokens;
export default new SuperTokens();
```

But then it's the same as using a class. There still is a need for checking if appInfo is undefined.

Did I miss a better way of importing/exporting singleton?

I checked in [supertokens-website](https://github.com/supertokens/supertokens-website/blob/master/lib/ts/index.ts#L143) and we are not doing singleton.

> The way to solve this is to have a separate structure of folders that will be used for imports only. The actual code will still be the way it's structured.

We can also fix by adding an "export" attribute to package.json (source [Stackoverflow](https://stackoverflow.com/questions/30302747/root-directory-in-package-json)). It is easier to understand than reorganising the structure of folders that are used for import. I updated with `exports` and tested in an example app.

.mocharc.yml
```diff
@@ -3,5 +3,5 @@ spec: test/**/*.test.js
 reporter: spec
 slow: 20000
 timeout: 30000
-require:
-  - isomorphic-fetch
\ No newline at end of file
+# require:
```
NkxxkN, line 6: Keep throwing 
```
✖ ERROR: Error: Cannot find module 'isomorphic-fetch'
```
I'm going to re-add and fix when starting to work on async calls.

NkxxkN:
- [x] You do not need recipeId and routes as static variables in EmailPassword
- [ ] `RECIPE_ID_GET_PARAM` => `RECIPE_ID_QUERY_PARAM` in upcoming commits. (I missed that one).

I think that covers all from above. Let me know if I missed some comments.

rishabhpoddar:
## Addressing your questions from the previous comment:

> How can we reset appInfo without such a function (without actually exposing the appInfo attribute) to make sure we do not expose a useless function?

You create a wrapper class for `SuperTokens` that is exposed to the user, and calls the functions of the actual implementation (which is in another file like `supertokens.ts`). In the wrapper class, we do not need to have the `reset` function. During testing, you can call the `reset` function by importing directly from `supertokens.ts`.

> We can also fix by adding an "export" attribute to package.json (source Stackoverflow). It is easier to understand than reorganising the structure of folders that are used for import. I updated with exports and tested in an example app.

Sounds good!!

## More feedback

- type `AppInfoBase` has `apiBasePath?: string` and `websiteBasePath?: string`, so does `AppInfoConfig`.. I'm a little confused. Perhaps `AppInfoBase` is not supposed to have those two items?
- The names `AppInfoConfig` and `AppInfo` are very confusing. Perhaps think of naming them `AppInfoUserInput` and `AppInfo`? Naming is very important. Please spend some time thinking about this next time :)
- In `package.json`, put `react` as a `peerDependency`. 
- In the `init` function, you are using an `if` statement without `{}`. Please be careful about this.
- If `init` is called twice, do we want to throw an error the second time, or just `return`?
- This line is very confusing for me: `return (SuperTokens.instance = new SuperTokens(config));`, and perhaps for anyone else too?
- `static` and `instance` variables should be defined on top of the class. So please move `private static instance?: SuperTokens;` to the top.
- For instance functions, please always use arrow functions. This way, we do not have to worry about binding
- In `normaliseUrl`, you do `url.startsWith("localhost:")`, what if the user gives `localhost` (which is = to `localhost:80`)?
- Does `normaliseUrl` throw an error in case the input has invalid syntax? In this case, we want to make that clear somehow.
- Won't `config.appInfo.apiBasePath !== undefined && normalisePath(config.appInfo.apiBasePath)` return a `boolean`?
- `config.appInfo.logoFullURL || undefined` === `config.appInfo.logoFullURL` since the type of `logoFullURL` is `string | undefined`
- Not sure if we need `getRecipeList` to be there as a static function or be exported to the user.
- As per the issues, `canHandleRoute` does not take a URL, and neither does `getRoutingComponent`
- Calling a static method from a non-static method is a little strange. For example, in `getRoutingComponent`, instead of doing `SuperTokens.getAppInfo()`, you can just do `this.appInfo`?
- Does this cover all the edge cases: `pathname.startsWith(SuperTokens.getAppInfo().websiteBasePath`?
- There seems to be a lot of conversation from `url string` to `URL object` within a function call. Perhaps you want to do that conversation at the top of a function stack trace and make all functions accept the normalised `URL object`?
- If you are using `this.recipeList.find(recipe => {`, then I prefer the entire block to be `pure`, cause its behaviour is not obvious to the reader. In your case, it's not pure, therefore, either make it pure, or convert it to a regular `for` loop.
- Please update the routing algorithm in [this issue](https://github.com/supertokens/supertokens-node/pull/23) as your own comment.
- I'm a little confused about the purpose of `FeatureHash` in a recipe. Why is it have the word "hash"?
- `return Boolean(this.features[routeWithoutBasePath]);` => i'm not sure what this does. Is it the same as `return this.features[routeWithoutBasePath] !== undefined`?

NkxxkN:
Not going to use package.json "exports" feature before https://github.com/microsoft/TypeScript/issues/33079 is solved.
It created many issues with no viable workaround, so going back to the first idea of having a separate structure of folders at the root that import files from `lib/build` until issue above is fixed.

package.json
```diff
@@ -34,11 +55,18 @@
...
   "license": "Apache-2.0",
   "bugs": {
     "url": "https://github.com/supertokens/supertokens-auth-react/issues"
   },
-  "homepage": "https://github.com/supertokens/supertokens-auth-react#readme"
+  "homepage": "https://github.com/supertokens/supertokens-auth-react#readme",
+  "dependencies": {
+    "@emotion/core": "^10.0.35",
+    "emotion": "^10.0.27",
+    "emotion-server": "^10.0.27",
```
NkxxkN, line 69: emotion server is a peer dependency of react-shadow if we want to use emotion with react-shadow.

package.json
```diff
@@ -2,18 +2,39 @@
...
+    "puppeteer": "^5.3.1",
+    "react": "^16.14.0",
+    "react-dom": "^16.0.0",
+    "react-router-dom": "^5.2.0",
+    "regenerator-runtime": "^0.13.7",
+    "typescript": "3.5.2"
+  },
+  "peerDependencies": {
+    "react": "^16.14.0",
+    "react-router-dom": "^5.2.0"
```
NkxxkN, line 29: peer dependency that is not required if you are not using the `getSuperTokensRoutesForReactRouterDom component`. Peer dependencies  versions discussed here https://github.com/supertokens/supertokens-auth-react/issues/14

NkxxkN:
This here https://github.com/supertokens/supertokens-auth-react/pull/11/files#diff-cb70d48ffe3b5c6df3235d8da1c91cfae355dc0233469677848b14567109b668R35 is the hardest to understand I suppose. I can walk you through the code on a call is necessary.

rishabhpoddar:
I have submitted a PR for this: https://github.com/supertokens/supertokens-auth-react/pull/15

NkxxkN:
I don't think we should add `AppInfo` to `RecipeModuleConfigs`. 

This make is mandatory to pass down AppInfo unrelated to the recipe module when it could just do `SuperTokens.getInstanceOrThrow().getWebsitePath()` or any other value. Since we can assume that SuperTokens will always be defined, it's better to use it as a single source of truth, and not propagate its configs down to its recipes.

rishabhpoddar:
We cannot call `SuperTokens.getInstanceOrThrow()` from the recipe's constructor since the contructor of SuperTokens has not yet finished.. So `SuperTokens.getInstanceOrThrow()` will throw an error. The other thing we can do is to initialise the recipes after the constructor of SuperTokens has finished.. but then that means that the recipeList is not properly initialised. So this implies that either it's an empty array by default or be `undefined` by default. It being an empty array is fine.. But I just prefer to pass the `appInfo` to the recipes.

NkxxkN:
I just realised that :/ 
Let's keep your solution giving `appInfo` to recipe module.

Issue #27, opened 2020-11-05
Code review for branch 0.0 - 5th Nov 2020


rishabhpoddar:
## Related to previous code review (https://github.com/supertokens/supertokens-auth-react/issues/20)
- [x] The `index.ts` of emailpassword does not does not export the components as part of the default export. Is that intentional? This means that users cannot do `<EmailPassword.SignInAndUp />`
- [x] Why do you have `export { SessionAPIWrapper };` in the `index.ts` of session recipe?
- [x] `getMatchingComponentForRouteAndRecipeId` should take path as NormalisedURLPath. Not as a string. You want to convert this class to a string as late as possible, and only if absolutely necessary.
- [x] The below 
   > (My comment) getMatchingComponentForRouteAndRecipeId is written in a way that if a rId is provided in the URL, but that doesn't match any recipe, then we will treat it as if rId is not there is not there in the URL. Is this expected?

   > (Your reply) "Yes that's what we decided in our previous discussion around routing. If no rid is provided, or if rid is unknown we want to use the first recipe feature that matches."

   Are you sure we had this conversation? Because on the backend, if the `rId` is there, but it not matching any route, then I pass the request to the user.

- [x] If we are allowing users to change the background for the default routes, then where will that go? If we allow them to do it via style in the init, what if they give a different colour for sign up and sign in?

- [x] The below
   > validateFormOrThrow in utils.ts does not throw any error.

   > It actually does throw if length is different as per our conversation.

   Yes, but the name suggests that if the form is valid, then it will not throw an error, else it will. Which means, as a consumer of this function I would not care about the return type.

- [x] It would be nice to move the API calls in SignInAndUp (and other components) to a different file that only contains API calls.

- [x] Does `fetch` throw an error in case the status code is >= 300?

## New issues
- [x] `onSuccessRedirectURL` in the config can be a full URL right (in case of multi-tenancy)? In that case, you do not want to normalise it with `NormalisedURLPath`. Perhaps there is no normalisation for this value that is required? (This applies to signInUp and to resetPassword feature)

- [x] When creating the `defaultSignInFields` in `utils.ts` (line 146), I think you forgot to add the email validator? If not that, you added the default password validator, overriding the user's password validator?

- [x] I feel the way the signIn form fields has been made can have bugs cause it's hard to understand for the following reasons (feel free to ignore this if you are very confident that your code regarding this is bug free):
   - Using reduce
[...]

Issue #26, opened 2020-11-04
Allow custom form fields values to be any type before calling the sign up API


rishabhpoddar:
Our theme right now supports only `string` values, but if someone makes their own theme, that may have other input values too. An input type of a number / boolean etc.. should be allowed in the form fields array when sending to the backend.

## Edit:
To allow for this on the backend, I had to make the following changes (pick whichever apply on the frontend):
- Changing types of the validate function to make the type of `value` from `string` to `any`
- Changing types of the handleCustomFields callback post sign up API (not relevant to the frontend)
- Changing default password and email validators to first check that the input is of type `string`

Here is the backend commit with the above changes for your reference: https://github.com/supertokens/supertokens-node/commit/20691af92a252282a73b8efbc15863ca3b4cc41a

PR #29, opened 2020-11-05
multiple fixes

NkxxkN:


NkxxkN:
Fix for https://github.com/supertokens/supertokens-auth-react/issues/26

NkxxkN:
Fix for https://github.com/supertokens/supertokens-auth-react/issues/30

NkxxkN:
- Fix https://github.com/supertokens/supertokens-auth-react/issues/31
- Fix `onCallSignUpAPI` and `onCallSignInAPI` props not implementing FDI https://github.com/supertokens/frontend-driver-interface/blob/master/v1.3.0.md#output-9 correctly

lib/.eslintrc.js
```diff
@@ -112,5 +112,14 @@ module.exports = {
...
             }
         ]
-    }
+    },
+    "overrides": [
+        {
+          // disable the rule specifically for user input validators files.
+          "files": ["**/validators.ts"],
+          "rules": {
+            "@typescript-eslint/explicit-module-boundary-types": ["off"]
```
NkxxkN, line 121: We want to allow `any` types for validators only.

NkxxkN:
Ready for review @rishabhpoddar

rishabhpoddar:
# IN PROGRESS
The review comments are above. A few more comments:
- I have not gone through `emailpassword/utils.ts` in too much detail. Just a skim. So please make sure the logic there is fine. One issue I think I found is that for password reset validators, whilst you use the password validator from sign up, you do not use the email validator from there (it's always equal to the default email validator, which is incorrect).

lib/ts/recipe/session/session.ts
```diff
@@ -125,6 +144,18 @@ export default class Session extends RecipeModule {
         return Session.getInstanceOrThrow().doesSessionExist();
     }
 
+    static addAxiosInterceptors(): boolean {
```
rishabhpoddar, line 147: This takes an `axios` instance. The function signature in `supertokens-website` is:
```
addAxiosInterceptors: (axiosInstance: any) => void
```

lib/ts/recipe/session/session.ts
```diff
@@ -125,6 +144,18 @@ export default class Session extends RecipeModule {
...
+        return Session.getInstanceOrThrow().addAxiosInterceptors();
+    }
+
+    static setAuth0API(): boolean {
```
rishabhpoddar, line 151: This also takes a param as in put.

lib/ts/recipe/session/session.ts
```diff
@@ -82,6 +89,18 @@ export default class Session extends RecipeModule {
...
 
+    addAxiosInterceptors = (): boolean => {
+        return this.sessionSdk.addAxiosInterceptors();
+    };
+
+    setAuth0API = (): boolean => {
+        return this.sessionSdk.setAuth0API();
+    };
+
+    getAuth0API = (): boolean => {
```
rishabhpoddar, line 100: Incorrect return type

lib/ts/recipe/session/session.ts
```diff
@@ -82,6 +89,18 @@ export default class Session extends RecipeModule {
         return this.sessionSdk.doesSessionExist();
     };
 
+    addAxiosInterceptors = (): boolean => {
+        return this.sessionSdk.addAxiosInterceptors();
+    };
+
+    setAuth0API = (): boolean => {
```
rishabhpoddar, line 96: Incorrect return type

lib/ts/recipe/session/session.ts
```diff
@@ -82,6 +89,18 @@ export default class Session extends RecipeModule {
         return this.sessionSdk.doesSessionExist();
     };
 
+    addAxiosInterceptors = (): boolean => {
```
rishabhpoddar, line 92: Incorrect return type

lib/ts/recipe/emailpassword/emailPassword.ts
```diff
@@ -158,26 +158,23 @@ export default class EmailPassword extends RecipeModule {
      */
 
     async signUpValidate(input: APIFormField[]): Promise<FormFieldError[]> {
-        return await validateFormOrThrow(input, this.config.signInAndUpFeature.signUpForm.formFields);
+        return await validateForm(input, this.config.signInAndUpFeature.signUpForm.formFields);
```
rishabhpoddar, line 161: This can throw an error in case the input fields do not have the same length as the config form-field. In this case, what will happen?

Ideally, you should stick to only throwing an error in a function from a function or not throwing one at all. But OK.

lib/ts/recipe/emailpassword/validators.ts
```diff
@@ -0,0 +1,95 @@
...
+
+    return undefined;
+}
+
+/*
+ * defaultLoginPasswordValidator.
+ * min 1 characters.
+ */
+
+export async function defaultLoginPasswordValidator(value: any): Promise<string | undefined> {
```
rishabhpoddar, line 77: On the backend, I do not have this kind of a function. For sign in, password validation, I simply return `undefined`. The core separately checks if the password is not an empty string and if it is, it throws the wrong credentials error.

Would you like to have the same behaviour here? Or shall I change the backend?

rishabhpoddar (changes requested):

rishabhpoddar:
The review comments are above. A few more comments:
- I have not gone through `emailpassword/utils.ts` in too much detail. Just a skim. So please make sure the logic there is fine. One issue I think I found is that for password reset validators, whilst you use the password validator from sign up, you do not use the email validator from there (it's always equal to the default email validator, which is incorrect).
- Please confirm that you have seen, applied and understood the unchecked item at the bottom [this comment](https://github.com/supertokens/supertokens-auth-react/issues/27#issuecomment-722900699) 
- The three new functions added in the session recipe are not being exposed via the `session/index.ts` file
- There are still occurrences of implicit checks with props in the Components. You do not need to change the code for this now since it probably won't be solved quickly.

lib/ts/recipe/emailpassword/components/signInAndUp/api.ts
```diff
@@ -0,0 +1,113 @@
...
+            return {
+                ...response
+            };
+        }
+
+        // Otherwise, something went wrong.
+        return {
+            status: API_RESPONSE_STATUS.GENERAL_ERROR,
+            message: SOMETHING_WENT_WRONG_ERROR
+        };
```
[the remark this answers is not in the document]
NkxxkN: Right, I wasn't sure if the API was returning `GENERAL_ERROR` in some cases or if it was only for status code > 300.
I've refactored so that API response type does not include GENERAL_ERROR and theme response types includes it.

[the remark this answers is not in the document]
rishabhpoddar: According to the FDI spec, there should never be a status of `GENERAL_ERROR` at all returned by the API.

lib/ts/recipe/emailpassword/validators.ts
NkxxkN: I was under the impression that we decided to at least verify that there was one character but I didn't find any traces of this conversation. 

So I decided to check what big corps do:
- Facebook: no check for empty password
- Twitter: Disable the login button if empty password
- Instagram: Disable the login button if password is less than 6 characters

There is already a verification to make sure that all required fields are not empty:
image

Which makes the  `value.length === 0` check useless.

You don't need to implement the same thing in the backend, as you said, you can simply throw wrong credentials.

NkxxkN:
> whilst you use the password validator from sign up, you do not use the email validator from there (it's always equal to the default email validator, which is incorrect).

Fixed

> Please confirm that you have seen, applied and understood the unchecked item at the bottom this comment

Done

> The three new functions added in the session recipe are not being exposed via the session/index.ts file

Fixed

> There are still occurrences of implicit checks with props in the Components. You do not need to change the code for this now since it probably won't be solved quickly.

Fixed

lib/ts/recipe/emailpassword/emailPassword.ts
NkxxkN: The app would crash in that case. That should only happen while developers are developing a custom theme, trying to submit an incomplete form to the backend during development.
This is fairly important that the custom theme is implemented correctly and that this error is not being ignored so I was thinking that it's fine to throw unexpectedly here. Let me know if that's ok for you

lib/ts/recipe/emailpassword/components/signInAndUp/SignInAndUp.tsx
```diff
@@ -180,67 +123,35 @@ class SignInAndUp extends Component<SignInAndUpProps, { isLoading: boolean; user
...
-            };
-        } catch (e) {
-            return {
-                response: {
-                    status: API_RESPONSE_STATUS.GENERAL_ERROR,
-                    message: "Something went wrong. Please try again"
-                }
+    setStateOnSuccessfulAPICall(normalisedAPIResponse: SignInThemeResponse | SignUpThemeResponse): void {
+        let user: User;
+        if (normalisedAPIResponse.status === API_RESPONSE_STATUS.OK && normalisedAPIResponse.user) {
```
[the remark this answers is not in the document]
rishabhpoddar: @NkxxkN this comment has not be resolved yet.

PR #34, opened 2020-11-12
Add StyleProvider and StyleConsumer

NkxxkN:
# Leverage the CreateContext React API to create a StyleProvider

## What it does

This method helps using styling defined by users from any child components without having to pass down the props to nested components. It reduces the effort of using user defined styles.

### Step 1

In `emailpassword/styles/stylesContext.tsx` we create a React context, and export both the provider and the consumer.
We feed the provider with an object containing  the "defaultStyles" and the "palette" as a value.

```tsx
const StyleContext = React.createContext({
    palette: defaultPalette,
    defaultStyles: getDefaultStyles(defaultPalette)
});

export function StyleProvider({ children }: { children: JSX.Element }): JSX.Element {
    const palette = EmailPassword.getInstanceOrThrow().getConfig().palette;

    return (
        <StyleContext.Provider
            value={{
                palette,
                defaultStyles: getDefaultStyles(palette)
            }}>
            {children}
        </StyleContext.Provider>
    );
}
```

### Step 2

In the base components (`SignInAndUp` feature and `ResetPasswordUsingToken` feature), wrap the rendered elements into the `StyleProvider` defined before.

```tsx
render = (): JSX.Element => {
        return (
            <StyleProvider>
                <FeatureWrapper useShadowDom={useShadowDom}>
                    // Component goes here.
                </FeatureWrapper>
            </StyleProvider>
        );
    };
```

### Step 3
In any child components that will need to access the defaultStyles and palette, wrap the rendered elements into the `ThemeConsumer` component that gives "defaultStyles" and "palette" as arguments.
 
Example of the library's `inputError` component:

```tsx
import { StyleConsumer } from "../../styles/styleContext";

export default function InputError({ style, error }: InputErrorProps): JSX.Element {
    return (
        <StyleConsumer>
            {({ defaultStyles }) => (
                <div className="inputErrorMessage" css={[defaultStyles.inputErrorMessage, style]}>
                    {error}
                </div>
            )}
        </StyleConsumer>
    );
}
```

## TODO

 - [ ] Update the docs in Detailed guide to use the StyleConsumer instead of using props in "making your own frontend"

### DIFF
--- a/lib/build/recipe/components/featureWrapper.d.ts
+++ b/lib/build/recipe/components/featureWrapper.d.ts
@@ -1,9 +1,7 @@
 /// <reference types="react" />
-import { NormalisedDefaultStyles } from "../emailpassword/types";
 declare type FeatureWrapperProps = {
-    defaultStyles: NormalisedDefaultStyles;
     children: JSX.Element;
     useShadowDom?: boolean;
 };
-export default function FeatureWrapper({ children, defaultStyles, useShadowDom }: FeatureWrapperProps): JSX.Element;
+export default function FeatureWrapper({ children, useShadowDom }: FeatureWrapperProps): JSX.Element;
 export {};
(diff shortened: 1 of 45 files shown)


NkxxkN:
fyi @rishabhpoddar

Issue #37, opened 2020-11-13
Sign up copy error


rishabhpoddar:
Below the sign up form, it is written:
"By signin up, you agree to ourTerms of Service and Privacy Policy"

- There is a typo in "signing"
- We want to show the terms of services and privacy policy only if they provide a link to those in the `init` function. If they provide a link to:
   - None: We do not want this sentence
   - Any one: We want just that mentioned in the sentence
      - "By signing up, you agree to our Terms of Service" or
      - "By signing up, you agree to our Privacy Policy"
   - Both:
      - "By signing up, you agree to our Terms of Service and Privacy Policy"

rishabhpoddar:
The link for the terms and privacy that the user gives must be as is I think. So if they give a path like `/hello`, then it will take them to that within their own site. However, if they give a url, then it will take them to that. 

The reason for this is that often the login is on something like `app.example.com`, but the privacy policy etc.. is on `example.com/privacy`.. So we need to allow them to give complete links. 

Right now, I tested by setting the links to `example.com`, however, it takes me to `localhost/example.com`

NkxxkN:
This is the expected behaviour of HTML `a` links. If you try with:
`<a href="example.com" >` on any websites it will redirect to `<CURRENT_SITE>/example.com`. You should provide a URL not a domain.

Renamed `termsAndConditionsLink` => `termsOfServiceLink` in both this package and docs.

rishabhpoddar:
I see. Perhaps we should create a normalised URL class that takes a url (domain + path) and normalises that.. so even if the usr gives `example.com/privacypolicy`, it behaves as expected?

NkxxkN:
I feel like trying to implement a URL parser on top of the one from HTML will add complexity and might be harder than we think resulting in other issues for other cases.
This is not important enough, the user can realise this themselves. 

Also most of the time, you go to your privacy page, copy the url and paste it in the configs which will work.

rishabhpoddar:
Hmm.. Okay. We can close this issue for now. However, I don't think it will be as complex since we have already made domain normlise and path normlise, which pretty much does most of the things. We can probably implement this by:

```
let searchQueries = extractSearchQueries(url)
urlNormlise(url) = domainNormalise(url) + pathNormalise(url) + searchQueries
```

PR #40, opened 2020-11-13
0.0 quick fixes

NkxxkN:
**privacy and terms links not defined:**

image

**privacy link only**
image

**terms link only**
image

**privacy and terms links**
image


NkxxkN:
Fix #37

Issue #44, opened 2020-11-13
Unable to call Session.addAxiosInterceptors(axios); as desired


rishabhpoddar:
We would like to call this function at the top of the file near the import statement. However, that means this may get executed before the `supertokens.init` function. In this case, it throws `Error: No instance of Function found. Make sure to call the "init" method.` error. A few points:
- [x] "No instance of Function found" is a bad error message
- [x] Make sure to call the "init" method. -> Make sure to call the "supertokens.init" method.
- [x] The error happens because in `addAxiosInterceptors`, we call the `getInstanceOrThrowError` function. Because the `init` function has not been called yet, an error is thrown. We can fix this by simply calling the `supertokens-website` functions from the static functions directly (for all the functions related to session).

PR #46, opened 2020-11-14
changes to session recipe

rishabhpoddar:
## Related issue:
- https://github.com/supertokens/supertokens-auth-react/issues/44

## Changes:
- Removed the instance function `addAxiosInterceptor`. Instead, we are calling the `sessionSdk` directly from the `static` function. This should fix the issue mentioned above.
- Some of the session function signatures were still incorrect. TS did not point it out because the type of `sessionSdk` was `any`. I have changed and fixed those.
- Maybe fixed incorrect statements in the `CONTRIBUTING.md` file 

## Testing
- I could not get the tests to run because of numerous errors while starting the various processes. 
- The biggest issue I faced was that even after `npm run test` failed, and I quit it, I was getting the message "Waiting for frontend...". This means there were some "orphan" processes running in the background even after I regained control of the shell.
- Please do test my changes if you can, or else help me setup the testing env on my computer please.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -25,7 +25,7 @@ We're so excited you're interested in helping with SuperTokens! We are happy to 
 4. Install the project dependencies
    ```
    npm i -d
-   cd react-test-app && npm i -d
+   cd ./test/react-test-app && npm i -d
    ```
 5. Add git pre-commit hooks
    ```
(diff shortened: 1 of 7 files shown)


NkxxkN:
No need for step 5 (`npm run server`) to run the tests. Simply run `npm run test`. Thanks for pointing it out. I've updated Contributing file.
That;'s probably the  reason why it kept failing for you. Though the behaviour you describe with "Waiting for frontend" still executing is a bit strange. It never happened to me.

NkxxkN (approved):

rishabhpoddar:
If all the tests are now passing, please feel free to merge, and close this.

Issue #48, opened 2020-11-14
Add tests for session recipe


rishabhpoddar:
You can do so by copying the tests on the `supertokens-website`. You need to run tests for `axios` and for `fetch`. Just adding one test for both should be enough since this SDK is a wrapper to the actual SDK. The one test can be a simple refresh test as done here:
- For axios: https://github.com/supertokens/supertokens-website/blob/master/test/axios.test.js#L261
- For fetch: https://github.com/supertokens/supertokens-website/blob/master/test/fetch.test.js#L223

PR #50, opened 2020-11-16
Fix 42 44

### LINKED ISSUE #42: Going to a route under a base path we control yields a blank screen
If I go to `/auth/hello`, I see a white screen. Instead, we should maybe redirect the user to `/`? Or to `/auth` (or whatever the base path is as specified by the user)

### LINKED ISSUE #44: Unable to call Session.addAxiosInterceptors(axios); as desired

NkxxkN:
Fix #42 

Fix #44 
  - Redirect to `websiteBasePath` which will show the login widget.
  - Modify `getSuperTokensRoutesForReactRouterDom` to:
       - render an array of `Route` components (not of components containing routes) => that fixes `<Route exact />`, hence allowing us to catch unknown routes correctly.
  - No need to sort routes anymore since `exact` behaves correctly now.

### DIFF
--- a/lib/build/components/superTokensRoute.js
+++ b/lib/build/components/superTokensRoute.js
@@ -72,24 +72,40 @@ function _interopRequireWildcard(obj) {
     return newObj;
 }
 
-/* Copyright (c) 2020, VRAI Labs and/or its affiliates. All rights reserved.
- *
- * This software is licensed under the Apache License, Version 2.0 (the
- * "License") as published by the Apache Software Foundation.
- *
- * You may not use this file except in compliance with the License. You may
- * obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
- * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
- * License for the specific language governing permissions and limitations
- * under the License.
- */
+function _toConsumableArray(arr) {
+    return _arrayWithoutHoles(arr) || _iterableToArray(arr) || _unsupportedIterableToArray(arr) || _nonIterableSpread();
+}
 
-/*
- * Imports.
- */
+function _nonIterableSpread() {
+    throw new TypeError(
+        "Invalid attempt to spread non-iterable instance.\nIn order to be iterable, non-array objects must have a [Symbol.iterator]() method."
+    );
+}
+
+function _unsupportedIterableToArray(o, minLen) {
+    if (!o) return;
+    if (typeof o === "string") return _arrayLikeToArray(o, minLen);
+    var n = Object.prototype.toString.call(o).slice(8, -1);
+    if (n === "Object" && o.constructor) n = o.constructor.name;
+    if (n === "Map" || n === "Set") return Array.from(o);
+    if (n === "Arguments" || /^(?:Ui|I)nt(?:8|16|32)(?:Clamped)?Array$/.test(n)) return _arrayLikeToArray(o, minLen);
+}
+
+function _iterableToArray(iter) {
+    if (typeof Symbol !== "undefined" && Symbol.iterator in Object(iter)) return Array.from(iter);
+}
+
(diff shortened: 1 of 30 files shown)


NkxxkN:
Note: 
IF `getSuperTokensRoutesForReactRouterDom` is on top of the `Switch` and IF the user sets the `websiteBasePath`  to `/` it will catch every routes that are unknown by ST and redirect to `/`. 
Would anyone want to do that? Should we prevent from doing such a thing?

rishabhpoddar:
> IF getSuperTokensRoutesForReactRouterDom is on top of the Switch and IF the user sets the websiteBasePath to / it will catch every routes that are unknown by ST and redirect to /.

This would be a case if someone has a domain dedicated to login like `login.example.com`

NkxxkN:
Got it! In that case that works as expected.

Now there is a difference in how routing is handled between using `react-router-dom` or custom router:
  - `/auth/unknown` will be redirected to `/auth` with `react-router-dom`
  - `/auth/unknown` will not be catch by `canHandleRoute` and will let the user decide`

We can:
  - Keep the implementation as it is.
  - Decide that the user needs to handle `/auth/unknown` himself. Usually with `react-router-dom` you will have a catch unknown route at the end of your `<Switch>` component to show a 404.
 - Decide that we catch `/auth/unknown` with custom router and redirect to `/auth`.

What do you think is best? @rishabhpoddar

rishabhpoddar:
I think it's best to keep both the behaviours the same. So the custom router should catch /auth/unknown and redirect to /auth.

NkxxkN:
What happens if the user wants to disable default implementation (remove routing) for Sign In and Sign Up.

It would be a bug if we still catch `/auth/unknown` and redirect to `/auth`. 

Routing is done regardless of which recipe is used, so we shouldn't have to verify that `EmailPassword` recipe is initialised and that `signInAndUpFeature.disableDefaultImplementation` config is false before redirecting. It doesn't seem right.

I think we shouldn't actually redirect at all but make sure we do not catch the routes (with both router methods), letting users display their 404 page. 

Does that sound good?

rishabhpoddar:
> I think we shouldn't actually redirect at all but make sure we do not catch the routes (with both router methods), letting users display their 404 page.

Ideally this is perfect! If we do not match a page (cause it's disabled by the user, or cause that path is something like `/auth/unknown`, we would like the user to be able to handle this. But I was under the impression that this is not possible with react router dom.. But if it is, then please go ahead.

NkxxkN:
I thought the `exact` keyword was not working but it was due to an issue with my implementation. 
The `Route` component were embedded in the `SuperTokensRouteWithRecipeId` component. I updated the implementation  so that  `Route` embeds `SuperTokensRouteWithRecipeId` and now `exact` works. So we are going to go with that solution.

rishabhpoddar:
Cool. Does the exact keyword get affected by thing like:
- `#...` at the end of the URL
- query params
- a trailing (or no trailing) slash at the end of the URL

NkxxkN:
- [x] Add e2e testing for sign in / sign up / reset password. 
Session is tested also because it is deeply integrated with EmailPassword:
   - tested interceptors are working correctly
   - using doesSessionExist from the SDK
   - The rest is just a wrapper around the SDK. 
   - We do not need to implement the tests mentioned in https://github.com/supertokens/supertokens-auth-react/issues/48 . Especially because they implement a custom login. 

> Cool. Does the exact keyword get affected by things like:
#... at the end of the URL
query params
a trailing (or no trailing) slash at the end of the URL

No it does not. For routing based on query params (like we do with rId), we are using the recommended way.

Issue #51, opened 2020-11-17
Error messages smooth animations


NkxxkN:
Similar to https://login.retool.com/auth/login for incorrect email.

Implement animations for default themes:
 - formFields error
 - general Error

PR #54, opened 2020-11-18
extract defaultStyles from features to isolate in themes

NkxxkN:
- There was an issue with normalising the palette and default styles in the Feature meant that default styles and palette would be propagated to any themes. This was fixed by not using the ProviderStyle in the features and not merging the palette with the default palette (from the default theme) in the feature. This will make it easier to implement new themes.

- Reorganise the folder structures so that:
    - components
    - components library
    - styles
    - consts
    - types 
That all apply to one theme are in the same folder.

- Added light animations for error field. And animation similar to Retool https://github.com/supertokens/supertokens-auth-react/issues/51 for general errors.

https://www.loom.com/share/d3cbb70ee9bf4eabad481ce2a382369a

- Update text for password recovery success

image

It seems like there is a lot changes but most of it is just moving components files and updating imports.
Only places worth taking a look in this review are:
  - Removing palette normalisation from EmailPassword feature normalisation https://github.com/supertokens/supertokens-auth-react/pull/54/files#diff-6592f46ba659f18cc589522ce5cd8027bee83b72e61016fbbc604b49a8aa9c94R65

 - Light error Fields animations: https://github.com/supertokens/supertokens-auth-react/pull/54/files#diff-be0a9df2fe3aae5a5412b74da5b81c7bd619c1a4048a99d56fbbafd9d623a97aR43
 - Merging of palette is now happening in styleProvider for default theme https://github.com/supertokens/supertokens-auth-react/pull/54/files#diff-f9171ba5661c505d92f7db4478ca1e0573071199b5dcd783b92588637e46117eR17


rishabhpoddar:
> Reduced the size of the container for login signup from 524px to 480px

Please undo this as reducing this will make it look out of proportion from a height / font size etc.. point of view.

PR #59, opened 2020-11-21
autocomplete email and password

NkxxkN:
Add autocomplete for Chrome/Safari/Opera
They do not all support new password but they support "email" and "current-password"

Firefox does not need any indication. Firefox finds out on its own the correct behaviour!


rishabhpoddar:
- I changed the base from master to 0.1
- I don't quite understand this feature. Will discuss over a call once before merging.

rishabhpoddar:
Please feel free to merge this :)

PR #80, opened 2020-11-28
0.2

rishabhpoddar:
# I have created this only for review purposes. 
- I have not reviewed the FormBase.tsx file since it's way too nuanced to review for someone who has not spent a long time working on it. So I trust you to double check the changes for clean state management and no silly mistakes. This holds true for many of the files that are JSX heavy as well (mostly to do with theme and library folder components).


lib/ts/recipe/emailpassword/emailPassword.ts
```diff
@@ -113,6 +114,21 @@ export default class EmailPassword extends RecipeModule {
         });
     };
 
+    verifyEmailExists = async (value: string, headers: HeadersInit): Promise<VerifyEmailAPIResponse> => {
```
rishabhpoddar, line 117: As per my previous comment, we should change the name of this function

lib/ts/recipe/emailpassword/types.ts
```diff
@@ -348,9 +360,19 @@ export type SignUpThemeProps = ThemeBaseProps & {
      * Call Sign Up API.
      */
     callAPI: (fields: APIFormField[]) => Promise<SignUpThemeResponse>;
+
+    /*
+     * Verify if an email exist.
+     */
+    validateEmail: (value: string) => Promise<string | undefined>;
```
rishabhpoddar, line 367: Why are we passing this as a prop to the theme? Shouldn't there be one function that is passed that takes the id and value of a field and returns `undefined | string`?

In the feature component, this function should then have an if around the id and pass the appropriate validator on it. In case of email, if should also check that the email does not exist.

lib/ts/recipe/emailpassword/types.ts
```diff
@@ -272,6 +282,8 @@ export type SignInAndUpProps = BaseProps & {
 
     onCallSignUpAPI?: (requestJson: RequestJson, headers: HeadersInit) => Promise<SignUpAPIResponse>;
 
+    verifyEmailExists?: (value: string, headers: HeadersInit) => Promise<VerifyEmailAPIResponse>;
```
rishabhpoddar, line 285: We should change the name to be consistent with others: `onCallEmailExistsAPI(email: string, header: ...)`

PR #81, opened 2020-11-28
rename verifyEmailExists, do not pass validateEmail to SignUp + remov…

NkxxkN:
…e usless states in SignIn/SignUp/EnterEmail/SubmitNewPassword

Follow up on https://github.com/supertokens/supertokens-auth-react/pull/80

# Explanation of how we verify if email exists on blur for sign up only:

- The normalisation of form fields provided by the user in init config is kept unchanged. We can't add the email validation API call here because we do not have access to a potential onCallEmailExistsAPI props provided by the user.

`SignInAndUp.tsx`, (The feature component), which is responsible for:
   -  API calls (built-in or provided by props)
   - Render the themes and give them props (Including formatted formFields).
In `SignInAndUp.tsx`, we add the optional `validateOnBlurOnly` attribute to the `props.signUpForm.formFields`. 

See => https://github.com/supertokens/supertokens-auth-react/pull/81/files#diff-6f8b309acf2aa215826cb1f2ced7a0332da2a044ee71d28fec9f222ca73eb8dcR259

That way we do not have to pass `validateEmail` as a props anymore to `SignUp` which was indeed not optimal.

Then, in FormBase, whenever a field is blurred, https://github.com/supertokens/supertokens-auth-react/blob/0.2/lib/ts/recipe/emailpassword/components/library/FormBase.tsx#L79

The choice of introducing this new optional validateOnBlurOnly field attribute makes is easier to add validation on blur only.
I thought about merging the `verifyEmail` to the validate method directly, but then how do you know the method is called in the context of input blur or if it is called in the context of a signup call?

### DIFF
--- a/lib/build/recipe/emailpassword/components/features/signInAndUp/SignInAndUp.d.ts
+++ b/lib/build/recipe/emailpassword/components/features/signInAndUp/SignInAndUp.d.ts
@@ -1,7 +1,7 @@
 import { PureComponent } from "react";
 import EmailPassword from "../../../emailPassword";
-import { SignInAndUpProps, SignInAndUpState, SignInThemeResponse, SignUpThemeResponse, OnHandleSignInAndUpSuccessContext, SignInAPIResponse, VerifyEmailAPIResponse } from "../../../types";
-import { APIFormField, RequestJson } from "../../../../../types";
+import { SignInAndUpProps, SignInAndUpState, SignInThemeResponse, SignUpThemeResponse, OnHandleSignInAndUpSuccessContext, SignInAPIResponse, VerifyEmailAPIResponse, FormFieldThemeProps } from "../../../types";
+import { APIFormField, NormalisedFormField, RequestJson } from "../../../../../types";
 import Session from "../../../../session/session";
 declare class SignInAndUp extends PureComponent<SignInAndUpProps, SignInAndUpState> {
     constructor(props: SignInAndUpProps);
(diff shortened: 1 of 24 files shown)


NkxxkN:
Updated docs https://github.com/supertokens/main-website/pull/77 for props `onCallEmailExistsAPI`

rishabhpoddar:
I don't think I am able to understand. We will have to discuss over the call. However, if validating is happening on blur, then why do we need to call validators again on sign up? Can't we keep the sign up button disabled while all validators have not succeeded?

Please feel free to merge this PR. But let's discuss this over a call. It will be easier.

NkxxkN:
> why do we need to call validators again on sign up?

Good question 🤔 

We do disable the SignUp button if all fields have not been validated.
The only exception is when only the currently focused field has not been validated yet. I realised that having to blur the last input in order to click on SignUp was a bad user experience. 

A potential optimisation would be to validate only this last input (find out which one is focused while clicking on the signup button and validate it).

rishabhpoddar:
Okay. As a note, even if we do not want to do the suggested optimisation before, and do all validation again on sign up, we should still call the check if email exists API when sign up is clicked. This is so that the code is cleaner. From a user experience point of view, it makes no difference at all.

NkxxkN:
> we should still call the check if email exists API when sign up is clicked

You are talking about verifying from the `POST /signup` call? Or calling `POST /signup/email/exists` first, then `POST /signup`?

PR #91, opened 2020-12-01
changes to add dev and release scripts as per supertokens/supertokens…

NkxxkN:
…-core#59


rishabhpoddar:
Seems good to me. Please feel free to merge. When you add a dev tag next, please check:
- A new commit is created with the dev tag and a message like "adding dev-vX.Y.Z tag to this commit to ensure building"
- The difference between the new commit and the previous one is nothing

PR #92, opened 2020-12-01
0.2 race conditions on blur

NkxxkN:
There was a race condition with updating FormBase status (hence, disabling the submit button).

Example of a failing scenario:

 - **User Action 1** Focus email, write email (ex: john@gmail.com)

 -  **User Action 2** Focus on password
     => As a result:
       - Start to run `onInputBlur` method for email (async call to verify if email exists)
       - Start to run `onInputFocus` method for password
       - Update state from `onInputFocus` method for password => Use provided field.id (=password) to detect that password is currently focused so leave the benefice of the doubt and enable Button (set state to "READY").
     - **Update state from `onInputBlur` method for email: Use field.id (=email) to detect that password is currently empty and not focused and disable Button  (set state to "IN_PROGRESS")**

 - **User Action 3** Write password, do not blur => button is still disabled

In order to fix this race condition issue:
 -  we are using the last up to date state from the setState method
 - we are setting the `isFocused` property directly on the field ref to make sure we always know the up-to-date focused input. That way, we do not rely on the provided field, which might not be focused anymore.

This fixes the issue for any asynchronous validator provided by any user.

### DIFF
--- a/lib/build/recipe/emailpassword/components/library/FormBase.d.ts
+++ b/lib/build/recipe/emailpassword/components/library/FormBase.d.ts
@@ -1,6 +1,6 @@
 import React, { PureComponent } from "react";
 import { APIFormField } from "../../../../types";
-import { FormBaseProps, FormBaseState, FormBaseStatus, FormFieldState } from "../../types";
+import { FormBaseProps, FormBaseState, FormFieldState } from "../../types";
 export default class FormBase extends PureComponent<FormBaseProps, FormBaseState> {
     static contextType: React.Context<{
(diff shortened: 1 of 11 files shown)


rishabhpoddar:
I don't really understand the description well, so in case you need my input, it's going to be difficult for me. But in case it's resolved, then no worries :)

NkxxkN:
It's a really weird race condition which took me a while to understand. I can explain over a call tmrw if you'd like but it's now fixed

rishabhpoddar:
Cool. I'm curious about anything weird.. especially if it's got to do with race conditions. So we can discuss tomorrow.

Issue #99, opened 2020-12-17
Shadow-root not supported on IE11


rishabhpoddar:
As pointed out in this comment: https://news.ycombinator.com/item?id=25459947

- [x] We should also do some research to see which other browsers we need to disable shadow-root for
- [x] In the docs, we need to mention this so that users who care about supporting those browsers, check for CSS clashes.

NkxxkN:
I confirm that it renders a blank page on IE11. It works on Edge.

Possible fix would be to detect browser and never use shadow dom for IE:

See https://stackoverflow.com/questions/9847580/how-to-detect-safari-chrome-ie-firefox-and-opera-browser

rishabhpoddar:
Yea. The suggested fix is good. Let's do that. Will add this issue to [the project](https://github.com/supertokens/supertokens-core/projects/5)

rishabhpoddar:
- [ ] We should also do some research to see which other browsers we need to disable shadow-root for
- [ ] In the docs, we need to mention this so that users who care about supporting those browsers, check for CSS clashes.

NkxxkN:
PR:

According to https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot, this is only IE.

According to [this article](https://blog.logrocket.com/should-you-support-internet-explorer/#:~:text=IE%2010%20is%20considered%20dead,compete%20with%20most%20modern%20browsers.), most modern app do not support Internet Explorer. 

Supporting Internet Explorer means a lot of extra efforts for less than 0.2% of people.

As a matter of fact, even `create-react-app` does not work with IE. You'd need to use https://www.npmjs.com/package/react-app-polyfill for retro-compatibility with IE. I tried that without success in our test app.

Now our lib should work with IE if the app supports IE (i.e. if they take care themselves of this polyfill fix).
I don't think there is anything extra we should do for IE support as I think most people agree that any modern web app will most likely fail to run on IE.

NkxxkN:
Is it even worth mentioning  IE in the docs?

rishabhpoddar:
> Is it even worth mentioning IE in the docs?

We can add a line about shadow-dom being disabled for IE in the place where we talk about the config param for disabling shadow-dom

NkxxkN:
https://github.com/supertokens/main-website/pull/171

PR #100, opened 2020-12-21
always disable shadow DOM for IE

NkxxkN:
# Disable shadow DOM for IE.

I've tested that it does not impact any other browser than IE.
I've tested that the code used to detect IE works properly (using https://www.browserling.com/).

I haven't successfully tested that the library works on IE11 since my test app does not work in IE11 (even without Supertokens). IE does not really work with react (see https://github.com/supertokens/supertokens-auth-react/issues/99).

### DIFF
--- a/lib/babel.config.js
+++ b/lib/babel.config.js
@@ -1,4 +1,4 @@
 module.exports = {
-    presets: ['@babel/preset-typescript', '@babel/preset-env', '@babel/preset-react'],
-    plugins: ["transform-class-properties"] 
-}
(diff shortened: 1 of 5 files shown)


rishabhpoddar:
I have seen the code, and it seems fine. Please feel free to merge whenever you want.

rishabhpoddar:
By the way, shouldn't you use `getWindow().document.documentMode` instead of `window.document.documentMode`?

NkxxkN:
Oh good catch! I'll see if I can create an eslint rule for that too

NkxxkN:
For reference: https://github.com/supertokens/supertokens-auth-react/pull/101

Issue #96, opened 2020-12-02
Button clicked visual output


NkxxkN:
We need to provide user feedback on button click on different forms

Related to https://github.com/supertokens/supertokens-auth-react/issues/45

NkxxkN:
See https://material-ui.com/components/buttons/#button

Or Facebook login

rishabhpoddar:
@mufassirkazi please have a look at this ans suggest what we should use for button click feedback

mufassirkazi:
https://zpl.io/2y4k99p

I suggest we can use the styling inside the buttons under "Sizes" heading in the link given by kevin. The buttom named "Custom CSS" under "Customised buttons" also will work. The colors to be used are given in the zeplin link at the start of this comment.

rishabhpoddar:
@NkxxkN I think we should do this soon. Having no feedback on button click is really odd.

NkxxkN:
Sure. I prioritized it in https://github.com/supertokens/supertokens-core/projects/5

PR #105, opened 2020-12-28
add ripple effect to buttons

NkxxkN:
ripple-button-effect

Introduces a new color `buttonFocus` color.

That could be a breaking change if someone changed their button color in `v0.2` and updates to `v0.3`.
We could prevent the breaking change by making this optional (and false by default) but:
  -  that would introduce a new config.
  -  people would not necessarily turn it on
  
@rishabhpoddar wdyt?

@mufassirkazi Slight changes from original designs because of technical complications, it is not exactly like the ripple effect from MaterialUI, but very similar. I didn't manage to get the exact same thing and I'd rather not spend another couple of hours since this result is satisfying enough. Let me know if that's ok with you.

Tested on Safari/Firefox/Chrome/Opera/Edge


NkxxkN:
https://github.com/supertokens/supertokens-auth-react/issues/96

rishabhpoddar:
- I think we should add a simpler effect.. One that is less noticeable, but still gives obvious feedback. In the new design, we can rethink this and perhaps make it a ripple effect.
- An idea for the buttonFocus colour could be to take the colour of the button, and make it a little darker (As opposed to hardcoding it). This way, even if the colour of the button is changed by the user, it will still work nicely, without an extra config. Is this possible @NkxxkN ?

NkxxkN:
Good idea @rishabhpoddar!

I explored with chroma-js and the result is pretty nice with all colors. Also feels a little bit less noticeable which I feel is better than the original colors.

**Note on choosing chroma (over hex-gba or color-convert):**
**Pros:**
- Ease of use:
   - Using chroma-js because it supports taking any types of colors as input (`blue | green | #ff0011 | rgb(255,0,0)`)  
   - Integrated "darken" method which is exactly what we wanted to do.
- 7k stars on github

**Cons:**
- Bundle is 13kb compiled


Ex:
**#ff9b33**
ripple-button-effect-default

**green**
ripple-button-effect-green

**rgb(255,0,0)**
ripple-button-effect-rgb

NkxxkN:
We can also make it even lighter

ripple-button-effect-lighter

rishabhpoddar:
This looks great! 
- So no extra config value right?
- What is the behaviour on mobile?

NkxxkN:
No extra configs.
On mobile you can see a colour change when you click but the effect is not really noticeable.

NkxxkN:
Concerning the version, It will actually be `3.0.0` since this is a visual change.

rishabhpoddar:
Why 3.0.0? The current version is 0.2.x

NkxxkN:
Excuse me, `0.3.0` :)

Issue #64, opened 2020-11-23
upgrade shadowDom and Emotion versions


NkxxkN:
Emotion v11 and shadowDom support for v11

NkxxkN:
+ Start using CacheProvider to fix issues with clashing libraries when used with chakra-ui for instance.

PR #106, opened 2020-12-29
Emotion cache key

NkxxkN:
# Upgrade to Emotion v11

## Issue

There is an issue when using `supertokens-auth-react` with other libraries (such as chakra-ui) which also uses Emotion.
They both share the same cache key and it results in css clashes.

## Solution

Upgrading to v11 allows two things:

 - **with shadow dom**: Upgrading to `react-shadow: 19.0.1` which uses emotion v11 and uses the Emotion v11 CacheProvider (With cache key `react-shadow`). The change is that now, styling is declared withinthe shadow Dom (As opposed to in the `head` html tag of the root HTML component as it was before). That means even if there are multiple components with shadowDom using react-shadow and emotion, they will not clash because they are isolated.
 
 
image

- **without shadow dom**: Upgrading to v11 allows us to define our cache key ourselves ("supertokens") which should be unique always, and never clash with other libraries.

image

In addition:
 - Stop importing google fonts in featureWrapper. 
 - Remove warning importing `withRouter` from `react-router`  (which is not a direct dependency) when it could be done from `react-router-dom`.


NkxxkN:
https://github.com/supertokens/supertokens-auth-react/issues/64

rishabhpoddar:
- Do you want to wait for the person who was having this error to use this branch and report if it fixed it for them? If yes, please notify that person to try it out whenever you are ready.
- Will this be a breaking change?
- If the version is X.Y.Z, which value are we going to increment for this change?

NkxxkN:
Yeah we are discussing with nxteworld on discord trying to solve his issue. I'll update the version. This is not breaking change this will be in `0.2.3`

Issue #102, opened 2020-12-22
EmailPassword Recipe: Email verification 


NkxxkN:
# EmailPassword Recipe: Email verification 

**Sub issue of https://github.com/supertokens/supertokens-core/issues/124**

## isEmailVerified

- `type isEmailVerified = () => Promise<boolean>`
- `import { isEmailVerified } from "supertokens-auth-react/recipe/emailpassword"`

- isEmailVerified will use an API for now. Later on, we can make it use sessions instead (as a caching layer).
Page loading time problem (in required mode) is solved by setting the default to email being verified and then querying to check if it has been. This makes sense since most of the time, in required mode, the is verified will be true.

- If the user visits the {websiteBasePath}/verify-email?rid=emailpassword and the email is already verified, then they will be redirected to the success callback.

## Configs

```
SuperTokens.init({
  recipes: [
     EmailPassword.init({
            emailVerificationFeature: {
                mode: "OFF" | "REQUIRED" | "OPTIONAL",
                disableDefaultImplementation: boolean,
                sendVerifyEmailScreen: {
                  style: {...}   
                },
                sendVerifyEmailBanner: {
                   style: {...}   
                },
                verifyEmailLinkClickedScreen: {
                   style: {...}  
                }
           }
        }
     })
  ]
```

## Default URLs
For required mode email verify UI: {websiteBasePath}/verify-email?rid=emailpassword
For email clicked UI: {websiteBasePath}/verify-email?rid=emailpassword&token=TOKEN

## Email verification (Required mode)

- User is logged in after sign up
- If email verification mode is "required" and email is not verified, show the "email verification screen" (Can resend the email from there)

### Auth Component
 
```js
import {EmailPasswordAuth} from "supertokens-auth-react/recipe/emailpassword";

<EmailPasswordAuth
  onCallIsEmailVerifiedAPI?: (headers: HeadersInit) => Promise<IsEmailVerifiedAPIResponse>
  doesSessionExist?: () => Promise<boolean>
  onHandleShowEmailVerificationScreen?: () => Promise<boolean> // if true, we do nothing, if false, we do default redirection.
>
  <LoggedInComponents />
</EmailPasswordAuth>
```

- `onCallEmailVerifiedAPI`: By default, use new FDI specs
- `doesSessionExist`: By default, use session recipe
- `onHandleShowEmailVerificationScreen`:
- Logic:
  - 1/ Use Session doesSessionExist
[...]

PR #110, opened 2021-01-06
0.4 email verification

NkxxkN:
Notes:
- Implementation of https://github.com/supertokens/supertokens-auth-react/issues/102
- Removing `generalErrorBackground` in favour of an automatically generated colour using chroma-js. (Needs docs update in v0.4)


NkxxkN:
Gif of new design + email verification in use in the demo application (Without the API yet)
email-verification

NkxxkN:
TODO: 
- [ ] Upgrade to `supertokens-node` version with email verification features
- [ ] Implement tests for email verification
- [ ] Update Changelog
- [ ] Update Docs with new designs screenshots
- [ ] Update "Quick Start" guide adding `EmailPasswordAuth` component
- [ ] Create "Common Customisation > Email Verification" docs.
     - [ ] About (Explain "OFF" | "REQUIRED" modes)
     - [ ] Embed in a page
- [ ] Create `auth-react` v0.4 docs with emailVerificationFeature configs.
- [ ] Add Dev Tag
- [ ]

rishabhpoddar:
There seems to be a tiny "glitch" before the email verification screen is shown.. can that be fixed?

NkxxkN:
> There seems to be a tiny "glitch" before the email verification screen is shown.. can that be fixed?

The glitch is the one that we identified and decided to accept in https://github.com/supertokens/supertokens-auth-react/issues/102#issue-772752857

> isEmailVerified will use an API for now. Later on, we can make it use sessions instead (as a caching layer).
> Page loading time problem (in required mode) is solved by setting the default to email being verified and then querying to check if it has been. This makes sense since most of the time, in required mode, the is verified will be true.

rishabhpoddar:
> The glitch is the one that we identified and decided to accept in

However, post sign up form, we should not have that glitch. Post sign up, we should check if the email is required, and then redirect the user to the /verify-email screen instead of the success callback.

NkxxkN:
> However, post sign up form, we should not have that glitch. Post sign up, we should check if the email is required, and then redirect the user to the /verify-email screen instead of the success callback.

Make sense!

lib/ts/recipe/emailpassword/components/features/emailVerification/emailVerification.tsx
```diff
@@ -0,0 +1,296 @@
...
+        return;
+    }
+
+    render = (): JSX.Element => {
+        // In case Email Verification Mode is not required, redirect to onSuccessRedirectURL.
+        if (
+            this.getRecipeInstanceOrThrow().getConfig().emailVerificationFeature.mode !==
+            EMAIL_VERIFICATION_MODE.REQUIRED
+        ) {
+            redirectToInApp(
```
rishabhpoddar, line 238: I don't think doing a redirect inside a render function is appropriate. Perhaps this should be done in componentDidMount

lib/ts/recipe/emailpassword/components/features/signInAndUp/SignInAndUp.tsx
```diff
@@ -211,10 +217,19 @@ class SignInAndUp extends PureComponent<SignInAndUpProps, SignInAndUpState> {
...
+        // redirect to email verification screen if sign up and email verification mode is required.
+        let onSuccessRedirectURL = this.getRecipeInstanceOrThrow().getConfig().signInAndUpFeature.onSuccessRedirectURL;
+        if (
+            context.action === SUCCESS_ACTION.SIGN_UP_COMPLETE &&
+            this.getRecipeInstanceOrThrow().getConfig().emailVerificationFeature.mode ===
+                EMAIL_VERIFICATION_MODE.REQUIRED
+        ) {
+            onSuccessRedirectURL = `${this.getRecipeInstanceOrThrow()
+                .getAppInfo()
+                .websiteBasePath.getAsStringDangerous()}${DEFAULT_VERIFY_EMAIL_PATH}?rid=emailpassword`;
```
rishabhpoddar, line 229: don't hardcode rid

lib/ts/recipe/emailpassword/components/themes/default/emailVerification/emailVerificationScreen/sendVerifyEmail.tsx
```diff
@@ -0,0 +1,134 @@
...
+        this.setState(() => ({
+            status: SEND_VERIFY_EMAIL_STATUS.SUCCESS
+        }));
+        if (this.props.onSuccess !== undefined) {
+            this.props.onSuccess();
+        }
+    };
+
+    sendEmail = async (): Promise<void> => {
+        const { callAPI } = this.props;
```
rishabhpoddar, line 60: Are you sure that `callAPI` is a good name for this? I don't think we have discussed the props given from the feature component to the theme component

lib/ts/recipe/emailpassword/components/themes/default/emailVerification/emailVerificationScreen/sendVerifyEmail.tsx
```diff
@@ -0,0 +1,134 @@
...
+        try {
+            const response = await callAPI();
+
+            if (response.status === API_RESPONSE_STATUS.OK) {
```
rishabhpoddar, line 64: Another API response is that the email has already been verified. That's not being handled.

lib/ts/recipe/emailpassword/components/themes/default/emailVerification/emailVerificationScreen/verifyEmailLinkClicked.tsx
```diff
@@ -0,0 +1,138 @@
...
+            this.props.onSuccess();
+        }
+    };
+
+    componentDidMount = async (): Promise<void> => {
+        const response = await this.props.callAPI();
+        this.setState(() => ({
+            status: response.status
+        }));
+        return;
```
rishabhpoddar, line 67: No need for return statement

lib/ts/recipe/emailpassword/components/themes/default/emailVerification/emailVerificationScreen/verifyEmailLinkClicked.tsx
```diff
@@ -0,0 +1,138 @@
...
+            status: VERIFY_EMAIL_LINK_CLICKED_STATUS.SUCCESSFUL
+        }));
+
+        if (this.props.onSuccess !== undefined) {
+            this.props.onSuccess();
+        }
+    };
+
+    componentDidMount = async (): Promise<void> => {
+        const response = await this.props.callAPI();
```
rishabhpoddar, line 63: How are 500 errors from the API handled?

lib/ts/recipe/emailpassword/components/themes/default/emailVerification/emailVerificationScreen/sendVerifyEmail.tsx
NkxxkN: Good catch. This should be really edge case as we first check in the feature if email is already verified before loading this page. https://github.com/supertokens/supertokens-auth-react/pull/115/files#diff-95106643fb53b192624f3131a1ca70f9276079cf6e9dd20517d11445e89d51d2R224

But if someone comes back to this page without reloading after they have validated their email, then we should simply redirect to onSuccessRedirect.

lib/ts/recipe/emailpassword/components/themes/default/emailVerification/emailVerificationScreen/sendVerifyEmail.tsx
NkxxkN: This is the name in use already for Sign Up / Sign In and reset password.
See https://github.com/supertokens/supertokens-auth-react/blob/42492bce51255148572f1b6742d7e72a00c2630b/lib/ts/recipe/emailpassword/types.ts#L340

lib/ts/recipe/emailpassword/components/features/emailVerification/emailVerification.tsx
```diff
@@ -0,0 +1,296 @@
...
+    };
+
+    redirectToVerifyEmailScreen = async (): Promise<void> => {
+        redirectToInApp(
+            `${this.getRecipeInstanceOrThrow()
+                .getAppInfo()
+                .websiteBasePath.getAsStringDangerous()}${DEFAULT_VERIFY_EMAIL_PATH}?rid=emailpassword`,
+            undefined,
+            undefined
+        ); // No history object provided here, we need to reload the page.
```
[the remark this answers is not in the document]
NkxxkN: The reason being because updating `/auth/verify-email?rid=emailpassword&token=TOKEN` to `/auth/verify-email?rid=emailpassword` will not reload the component. 

There might be a way to leverage the `EmailVerification` feature state to reload the component while updating the URL too.

I'll try that.

lib/ts/recipe/emailpassword/components/features/emailVerification/api.ts
```diff
@@ -0,0 +1,103 @@
...
+        };
+        const response = await sendVerifyEmailAPI(headers);
+
+        // Otherwise, success.
+        if (response.status === API_RESPONSE_STATUS.OK) {
+            return {
+                status: API_RESPONSE_STATUS.OK
+            };
+        }
+
```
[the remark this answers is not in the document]
NkxxkN: Done in next PR

lib/ts/recipe/emailpassword/components/features/emailVerification/emailVerification.tsx
```diff
@@ -0,0 +1,296 @@
...
+    onHandleSuccess = async (context: onHandleEmailVerificationSuccessContext): Promise<void> => {
+        // If props provided by user, and successfully handled.
+        if (this.props.onHandleSuccess !== undefined) {
+            await this.props.onHandleSuccess(context);
+        }
+
+        // Otherwise, do nothing.
+    };
+
+    onSignInClicked = (): void => {
```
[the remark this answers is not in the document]
NkxxkN: We don't need it. I'm removing.

lib/ts/recipe/emailpassword/components/themes/default/emailVerification/emailVerificationScreen/verifyEmailLinkClicked.tsx
NkxxkN: They are handled in `emailVerification/api.ts`. No need to handle here. 

`sendVerifyEmail` was handling it but it was useless so I'm cleaning.

lib/ts/recipe/emailpassword/components/features/emailVerification/emailVerification.tsx
```diff
@@ -0,0 +1,296 @@
...
+        );
+    };
+
+    onEmailVerifiedSuccess = async (): Promise<void> => {
+        await this.onHandleSuccess({
+            action: SUCCESS_ACTION.EMAIL_VERIFIED_SUCCESSFUL
+        });
+    };
+
+    sendVerifyEmail = async (): Promise<SendVerifyEmailThemeResponse> => {
```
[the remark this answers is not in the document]
NkxxkN: `sendVerifyEmail` is used in the `componentDidMount` method and in the theme so it's useful to have a wrapper for it. `handleVerifyEmailAPI` was created to extract the API logic to  new files to reduce the size of each features.

The only place where I decided to not create an `api.ts` file is for `EmailPasswordAuth` `isEmailVerifiedAPI` because the wrapper code is really tiny: 

```
    isEmailVerifiedAPI = async (): Promise<boolean> => {
        try {
            const response = await this.onCallIsEmailVerifiedAPI({
                rid: this.getRecipeInstanceOrThrow().getRecipeId()
            });
            return response.isVerified;
        } catch (e) {
            // In case of API failure, continue, do not break the application.
            return false;
        }
    };
    ```

PR #115, opened 2021-01-08
0.4 send verify email on page load

NkxxkN:


rishabhpoddar:
I think you have missed one change in `.../components/themes/default/emailVerification/emailVerificationScreen/sendVerifyEmail.tsx` where where you call `callAPI`, you do not handle the case for email already verified. The comment for this is in the review for the other PR here: https://github.com/supertokens/supertokens-auth-react/pull/110#discussion_r555806491

Issue #119, opened 2021-01-13
Add preAPIHook, deprecate all onCallAPI methods


NkxxkN:
Sub issue of https://github.com/supertokens/supertokens-auth-react/issues/117

## Add 

```
type PreAPIHookContext = { 
  action: "SEND_RESET_PASSWORD_EMAIL" | "EMAIL_EXISTS" | "SUBMIT_NEW_PASSWORD" | "VERIFY_EMAIL" | "SEND_VERIFY_EMAIL" | "IS_EMAIL_VERIFIED" | "SIGN_IN" | "SIGN_UP" | "SIGN_OUT",
  requestInit: RequestInit
}

SuperTokens.init({
 appInfo: {...},
 recipeList: [
     EmailPassword.init({
           preAPIHook(context: PreAPIHookContext): Promise<RequestInit> {
               // Example:
               switch(context.action) {
                  case "SIGN_UP":
                    context.requestInit.headers["new-user"] = true;
                    break;
                  case "SIGN_IN":
                    // do things...
                    break;
                  default:
                    break;
                }
                return context.requestInit;
           }
       })
   ]
});
```

## Remove

**SignInAndUp:**
 - onCallSignUpAPI
 - onCallSignInAPI
 - onCallEmailExistsAPI
 
**ResetPasswordUsingToken**:
  - onCallSubmitNewPasswordAPI
  - onCallSendResetEmailAPI
  
 **EmailVerification**:
   - onCallSendVerifyEmailAPI
   - onCallVerifyEmailAPI
    
**EmailPasswordAuth**:
  - onCallIsEmailVerifiedAPI

PR #123, opened 2021-01-13
0.4 pre api hook

NkxxkN:
Fix https://github.com/supertokens/supertokens-auth-react/issues/119

- Make sure that `isEmailverified` returns true on API failure to prevent disruption of users with verified email.
- Update tests to remove custom API calls and leverage `preAPIHooks`


rishabhpoddar:
Let's discuss this over a call.

PR #126, opened 2021-01-13
0.4 onRedirection + onHandleEvent

NkxxkN:
Contains commits from https://github.com/supertokens/supertokens-auth-react/pull/124 and https://github.com/supertokens/supertokens-auth-react/pull/123 so merging all together.


rishabhpoddar:
These are a lot of code changes. Let's go through them at a high level over a call.

PR #132, opened 2021-01-18
add right padding to input

NkxxkN:
Fix https://github.com/supertokens/supertokens-core/issues/162

auth-react-inpout-right-padding


rishabhpoddar:
This is applied to all text inputs right?

NkxxkN:
Yes

PR #138, opened 2021-01-22
0.5.input adornments

NkxxkN:
The issue with having "show password" and "Success/Error" tick in the input is that we need to create input wrapper:

```js
<div class="inputWrapper">
   <input class="input">
   <showPassword/>
   <SuccessOrErrorIcon/>
</div>
```

The input wrapper gets all the styling like if it was the actual input (border and background colour).

**But the issue is that the browser autofill would do something ugly which is why I reverted in previous versions.** 

Chrome:

Firefox:
image
image

**But I finally managed to remove all browser styling for Firefox/Chrome/Safari and Edge!**

That means we can get back error/success icons and show password icon back in the inputs!

### With shadow dom enabled (no password manager)
ST_INPUT_SHADOW

### Without shadow dom enabled (with password manager)
ST_INPUT_NO_SHADOW

Note 1 :
Alignment of icons is off in the above gif. It was updated after.

image

Note 2:
Show password goes on the right of the success/error tick because "Show Password" is clickable but the success/error tick disappears on focus. That means when there is no focus, the show password would slide to the right resulting in clicking on the input... So we make sure the show password does not move on click by putting it to the right of the input wrapper.

Note 3: 
Introducing a `textInput` colour as the colour of input was always black, which i didn't realise could be an issue (see the example of the dark theme).

image

Note 4:

Designs have been discussed and validated by @mufassirkazi


rishabhpoddar:
- Has this been checked on mobile devices?
- This will require doc changes (new version + new user facing config changes)
- Please feel free to merge this PR.

NkxxkN:
Checked on mobile.
Yes I'm going to update the screenshots + add the input colour.

Issue #144, opened 2021-01-25
Add a new example with Redwood JS


rishabhpoddar:

NkxxkN:
Moving the conversation from the PR to this issue:

Not sure how to connect SuperTokens Node middleware system with Redwood functions as there does not seem to be any req, res objects accessible.

We might have to bypass all the logic using request, response, and return an object instead of sending the response?

rishabhpoddar:
Closing in favour of https://github.com/supertokens/supertokens-core/issues/129

PR #147, opened 2021-01-25
0.5 with redwood

NkxxkN:
## In Progress

Not sure how to connect SuperTokens Node middleware system with [Redwood functions](https://redwoodjs.com/docs/serverless-functions?search=services) as there does not seem to be any `req, res` objects accessible. 

We might have to bypass all the logic using `request, response`, and return an object instead of sending the response?


NkxxkN:
https://github.com/supertokens/supertokens-auth-react/issues/144

NkxxkN:
Closing for now. Issue #144 still open.

toxsick:
> Closing for now. Issue #144 still open.

This looked promising. So right now there is no good way to do this due to missing `req, res` objects in redwood?

regards

NkxxkN:
Hey @toxsick, thanks for your interest :) 

This was a quick attempt trying to figure out what is missing to get the integration working smooth. We got some good learnings from it.
The issue is still open but this draft PR is not needed anymore. We will definitely find a way to make it work. It will require changes to the libs directly though which needs to be planned in advance.

toxsick:
Hey @NkxxkN, thanks for the clarification. I am looking forward to give it a try when it becomes available.

Thanks for your work and have a nice one!

Issue #77, opened 2020-11-27
Implement `redirect_to` feature


NkxxkN:
# Feature Proposal

Implement `redirect_to` feature on successful login

## Flow

- Logged out user goes to `/app/loggedin`.
- EmailPasswordAuth catches it and informs `EmailPassword` recipe manager that onSuccessRedirectURL should be updated with `/app/loggedin` (Instead of updating `onSuccessRedirectURL`, we could have a function that first checks if value provided by `EmailPasswordAuth` is present **and valid**, otherwise, uses onSuccessRedirectURL)
- On successful login, `SignInAndUp` feature uses previsously defined function to redirect to `/app/loggedin`.

## Implementation

In `EmailPasswordAuth` component:
 - Use `Session.doesSessionExist()`
   - if false => redirect to `/auth`. Send current pathname to EmailPassword instance.
   - if true render children

In Email Password Feature, on successful login, call `getOnSuccessRedirectUrl()`
  - if temporary redirect URL is present redirect to  it, otherwise, redirect to `onSuccessRedirectURL`

## Follow up

- Use in https://github.com/supertokens/supertokens-demo-react
- Create a "EmailPassword > Common customisations > Sign In > Redirect To" doc section to explain how to use it.

## Security

- Make sure to not allow for open redirects => https://cheatsheetseries.owasp.org/cheatsheets/Unvalidated_Redirects_and_Forwards_Cheat_Sheet.html

Use `normalisedURLPath` on the provided path.

Since we use `getWindowOrThrow().location.pathname` to populate the temporary redirect URL there should be no open redirect. Also, since there is no `?redirect=` args, there should be no user manipulation possible, making it impossible to leverage any potential open redirect.

## Tests
 
- No open redirects: Test against: https://github.com/payloadbox/open-redirect-payload-list#open-redirect-payload-list-
- Redirect to previous path if present, otherwise, if `/auth` was accessed directly, redirect to `onSuccessRedirect`.

PR #150, opened 2021-01-26
Redirect to

NkxxkN:
- Implement redirect to
- Fix SPA reload issue when redirected by `EmailPasswordAutth`


rishabhpoddar:
- Can you please elaborate more on what `Implement redirect to` means?
- Also, does `Fix SPA reload issue when redirected by EmailPasswordAutth` mean that it was reloading the page instead of doing the SPA friendly thing of pushing the history?

NkxxkN:
> Can you please elaborate more on what Implement redirect to means?

Forgot to link the issue https://github.com/supertokens/supertokens-auth-react/issues/77
Scenario is:
 - Base redirection is `/dashboard`
- User goes to /app/custom/path` which is protected by `<EmailPasswordAuth>`
- No session => Redirect to `/auth`
- User authenticates and gets redirected to `/app/custom/auth` instead of usual `/dashboard`

> Also, does Fix SPA reload issue when redirected by EmailPasswordAuth mean that it was reloading the page instead of doing the SPA friendly thing of pushing the history?

Yes, the issue is that `EmailPasswordAuth` was not given any `history` props attribute because it was not wrapped with  `react-router-dom` `withRouter` method.

rishabhpoddar:
A few questions:
- How is the location of the redirection signalled to the auth component post successful sign in / up? For example, via query params in the URL?
- If the redirection url (due to user navigation) is not the same as what is given in the onSuccessURL config, then what happens?
- If the redirection url (due to user navigation) is there, will the success callback (I think it's called `onRedirect` function) be called? If yes, and it returns `false`, will the user be redirected to where they were? Should we pass the redirection URL in the success callback handler?
- This will probably require doc changes?

NkxxkN:
> How is the location of the redirection signalled to the auth component post successful sign in / up? For example, via query params in the URL?

The `EmailPasswordAuth` store the current `window.location.pathname` in the `EmailPassword` instance before redirecting to `/auth`. On successful `signIn/signUp`, we redirect to this value if it exists and clean it up.

The advantage of this approach is that we do not need to care about users tempering with the `redirect` attribute from the URL. 
The downside is that it relies on SPA not being reloaded, which is actually the case only with `react-router-dom`.
So this implementation will not work for nextJS and blitzJS.
Should we store that information in the sessionStorage instead? 


> If the redirection url (due to user navigation) is not the same as what is given in the onSuccessURL config, then what happens?


> If the redirection url (due to user navigation) is there, will the success callback (I think it's called onRedirect function) be called? If yes, and it returns false, will the user be redirected to where they were? Should we pass the redirection URL in the success callback handler?

In that case we do not even call the user provided callback. This takes precedence on the user callback.

> This will probably require doc changes?

We can mention it in the EmailPasswordAuth auth-react page. We can also mention this in the `getRedirectionURL` callback.
I don't think this is a major change and the logic is pretty easily understandable for customer.

rishabhpoddar:
> So this implementation will not work for nextJS and blitzJS.

That's an issue. We need to work always. It's a useful feature imo.

> Should we store that information in the sessionStorage instead?

No. Other providers add a redirectURI to the query params for the page. This is not a security issue as on redirect, the other page does not get any tokens etc.. it's a simple redirect.

> In that case we do not even call the user provided callback. This takes precedence on the user callback.

I think this is not good. One callback (the `handleEvent`) one should be called. Second, the redirect callback should be called, with the redirect URL as a param to it (by default, the redirectURL will be what the user gives in the config). If we are making this change, then the docs will have to change too.

rishabhpoddar:
Based on our discussion:
- We have the redirectToPath query params when navigating to the /auth page. We do not use sessionstorage since we want to support multi tenancy. 
- The value `redirectToPath` must be a path only, and not a full URL. This is to prevent open redirect attacks. We must enforce that it's just a path
- The `getRedirectionURL` takes the redirectToPath path (in case the context.action === "SUCCESS"), so that the user can use that if they like. The path can be accessed via context.redirectToPath
- When redirecting, the priority is to call the `getRedirectionURL` first (make sure we do not prepend the hostname to the returned value), then to use the `redirectToPath` path, then to use the config value.

lib/ts/recipe/recipeModule.ts
```diff
@@ -73,38 +73,32 @@ export default abstract class RecipeModule {
...
-            return await redirectToInApp(redirectUrl, title, history);
+            return await redirectToInApp(redirectUrl, history, queryParams);
         }
     };
 
-    getRedirectionURL = async (context: { action: string }): Promise<string> => {
-        let redirectUrl;
+    getRedirectionURL = async (context: { action: string; redirectToPath?: string }): Promise<string> => {
         // If getRedirectionURL provided by user.
         const getRedirectionURL = this.hooks.getRedirectionURL;
```
rishabhpoddar, line 93: I prefer if this is normalised in the utils just like all other input configs. So in utils, we check if the user has provided this function, if not, we set the config to have the default function. This way, when using, we can directly call the function via config without worrying about undefined checks and without having this extra function here.

lib/ts/recipe/recipeModule.ts
```diff
@@ -73,38 +73,32 @@ export default abstract class RecipeModule {
     };
 
     redirect = async (
```
rishabhpoddar, line 75: Le's discuss the code structure for redirect. It's a little hard to understand at the moment.

lib/ts/recipe/emailpassword/components/features/emailVerification/emailVerification.tsx
```diff
@@ -128,7 +151,10 @@ class EmailVerification extends PureComponent<FeatureBaseProps, { token: string
                     action: EMAIL_PASSWORD_SUCCESS_ACTION.VERIFY_EMAIL_SENT
                 }),
             onEmailAlreadyVerified: () =>
-                this.getRecipeInstanceOrThrow().redirect({ action: "SUCCESS" }, this.props.history)
+                this.getRecipeInstanceOrThrow().redirect(
+                    { action: EMAIL_PASSWORD_REDIRECTION_URL_ACTION.SUCCESS },
```
rishabhpoddar, line 155: I prefer not using constants like `EMAIL_PASSWORD_REDIRECTION_URL_ACTION.SUCCESS`. Adds more boilerplate coding etc.. instead make action have a specific type based on string and use that everywhere.. is easier to read, code etc.

lib/ts/recipe/emailpassword/components/emailPasswordAuth.tsx
```diff
@@ -109,4 +115,11 @@ class EmailPasswordAuth extends PureComponent<FeatureBaseProps, EmailPasswordAut
     };
 }
 
-export default EmailPasswordAuth;
+export default (function(): ReactComponentClass {
+    try {
```
rishabhpoddar, line 119: There are several cases of `try { ... } catch() { ... }`. This pattern in general indicates a lack of clear understanding of how the code in the try catch works. Whilst it's not a big problem, it's generally not the best coding practice.

rishabhpoddar (changes requested):

lib/ts/recipe/emailpassword/components/emailPasswordAuth.tsx
NkxxkN: We need try catch to wrap `react-router-dom` optional dependency here.

lib/ts/types.ts
```diff
@@ -116,7 +116,7 @@ export type RecipeModuleHooks = {
     /*
      * Optional method used for redirections.
      */
-    getRedirectionURL?: (context: { action: string }) => Promise<string | undefined>;
+    getRedirectionURL?: (context: { action: string; redirectToPath?: string }) => Promise<string | undefined>;
```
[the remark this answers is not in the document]
NkxxkN: I don't like how recipe module is typed either but since it's generic. We have no idea if action is "SUCCESS", "ERROR", etc... it depends on the recipe. That's one reason I wasn't sure we should have `redirectToPath` as a query params, we can discuss to try to find a better solution.

rishabhpoddar:
@NkxxkN the redirectTo param is also being sent to `verify-email` path, which is not needed since that page doesn't actually go anywhere.. So please change the package to only add that param in case a session does not exist.

PR #153, opened 2021-01-28
prune TS unused types and methods

NkxxkN:
Detect unused methods and types. 
Add to precommit hook to make sure we never release unused methods intentionally.


rishabhpoddar:
This change to pre commit is useful for the node package too. Please feel free to add it to that as well.

PR #157, opened 2021-02-02
With one login many subdomains

Infi-Knight:
auth.example.com -> a.example.com and b.example.com

Related to https://github.com/supertokens/main-website/issues/243


NkxxkN:
Discussed with Ravi and Joel. 

It seems in this use case we might need two separate applications:

## Application 1: Running on `auth.example.com`:

```js
SuperTokens.init({
  appInfo: {
    appName: "SuperTokens Demo App",
    apiDomain: getApiDomain(),
    websiteDomain: "auth.example.com",
    websiteBasePath: "/" // Change to have login on the base path of the domain.
  },
  recipeList: [
    EmailPassword.init({
      getRedirectionURL(context) {
        if (context.action === "SUCCESS") {
          // redirect to correct user subdomain.
        }
      }
```


## Application 2: SaaS application:
```js

const websiteDomain = window.location.origin // example: supertokens.example.com
SuperTokens.init({
  appInfo: {
    appName: "SuperTokens Demo App",
    apiDomain: getApiDomain(),
    websiteDomain
  },
  recipeList: [
    EmailPassword.init({
      signInAndUpFeature: {
        disableDefaultImplementation: true
      },
      emailVerificationFeature: {
        disableDefaultImplementation: true
      },
      resetPasswordUsingTokenFeature: {
        disableDefaultImplementation: true
     }
      getRedirectionURL(context) {
        if (context.action === "SIGN_IN_AND_UP") {
           return "http://auth.example.com/"
        }
         if (context.action === "RESET_PASSWORD") {
           return "http://auth.example.com/reset-password"
        }
        if (context.action === "VERIFY_EMAIL") {
           return "http://auth.example.com/verify-email"
        }
      }
   }]
}
```

NkxxkN:
The option of having only application would not really make sense. In that case, why would you have a subdomain for authentication only?

@rishabhpoddar let us know your thoughts on that one.

rishabhpoddar:
> The option of having only application would not really make sense

I don't think I understand what this sentence is trying to say.

-------
Why do we need two applications? Why can't we call init based on the current domain? (Since the current domain is different, it's not an SPA anyway)

NkxxkN:
We could call `init` based on the current domain, in that case why would you even bother redirecting to a `auth.example.com` domain if it does exactly the same as `supertokens.example.com`? It just does not make sense to implement such a redirection.

examples/with-one-login-many-subdomains/api-server.js
```diff
@@ -13,18 +13,13 @@ const websitePort = process.env.REACT_APP_WEBSITE_PORT || 3000;
...
   process.env.REACT_APP_WEBSITE_URL || `http://auth.example.com:${websitePort}`;
 
-let whiteList = [];
-whiteList.push(websiteDomain);
-whiteList.push(`http://abc.example.com:${websitePort}`);
-whiteList.push(`http://xyz.example.com:${websitePort}`);
-whiteList.push(`http://qually.example.com:${websitePort}`);
-whiteList.push(`http://gmail.example.com:${websitePort}`);
-
+const domainWhiteList = /http?:\/\/([a-z0-9]+[.])example[.]com/
```
NkxxkN, line 16: Same comment as https://github.com/supertokens/supertokens-auth-react/pull/158/files/0aa6f2365705daacedbe3e76df1494bb6ddf9f18..bf88748ff84654cf8ae19cabc16bb734ece8f353#r569353718

rishabhpoddar:
> We could call init based on the current domain, in that case why would you even bother redirecting to a auth.example.com domain if it does exactly the same as supertokens.example.com? It just does not make sense to implement such a redirection.

In this setup, the auth domain is auth.example.com, which redirects the user to their own subdomain. `example.com` on the other hand is like a homepage for that site, and there is no `supertokens.example.com`.

@NkxxkN @Infi-Knight does the above scenario make sense?

examples/with-one-login-many-subdomains/frontend/src/App.js
```diff
@@ -79,7 +101,7 @@ function App() {
             ) : (
               <Route path="/">
                 <EmailPassword.EmailPasswordAuth>
-                  <Home />
+                  {isUserOnValidSubdomain ? <Home /> : <Spinner />}
```
NkxxkN: Time to first render is a pretty crucial metric in SPAs. Adding a backend call before we even render anything simply to check if the user is on the correct subdomain seems like a bad trade-off to me. 

We have such a pattern already for email verification. We decided that it was better to:
 - render the application assuming that email is verified. 
 - verify if that's the case in the useEffect hook.
 - Do nothing if true. Redirect if false.

NkxxkN (approved):
Looks good to me! 
Great job on your first PR to this project 💪

Infi-Knight:
> Looks good to me!
> Great job on your first PR to this project



> Looks good to me!
> Great job on your first PR to this project

Thanks a lot :)

rishabhpoddar:
@NkxxkN if your review comments are taken care of, please resolve them.

PR #164, opened 2021-02-04
redirect using websiteBasePath comparison with current origin for mul…

NkxxkN:
This will help for multi tenancy.

### DIFF
--- a/lib/ts/recipe/recipeModule.ts
+++ b/lib/ts/recipe/recipeModule.ts
@@ -76,6 +76,14 @@ export default abstract class RecipeModule {
             new URL(redirectUrl); // If full URL, no error thrown, skip in app redirection.
         } catch (e) {
             // If history was provided, use to redirect without reloading.
+
+            // For multi tenancy, If mismatch between websiteDomain and current location, prepand URL relative path with websiteDomain.
+            if (getWindowOrThrow().location.origin !== this.appInfo.websiteDomain.getAsStringDangerous()) {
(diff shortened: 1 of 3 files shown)


lib/ts/recipe/recipeModule.ts
```diff
@@ -76,6 +76,14 @@ export default abstract class RecipeModule {
             new URL(redirectUrl); // If full URL, no error thrown, skip in app redirection.
         } catch (e) {
             // If history was provided, use to redirect without reloading.
+
+            // For multi tenancy, If mismatch between websiteDomain and current location, prepand URL relative path with websiteDomain.
+            if (getWindowOrThrow().location.origin !== this.appInfo.websiteDomain.getAsStringDangerous()) {
```
rishabhpoddar, line 81: - How does `.origin` behave with a port being specified in the current URL? Because I think `websiteDomain.getAsStringDangerous()` does return the port as well.

- Perhaps you should do new NormalisedURLDomain(getWindowOrThrow().location.origin). getAsStringDangerous() !== this.appInfo.websiteDomain.getAsStringDangerous()

rishabhpoddar (changes requested):

lib/ts/recipe/recipeModule.ts
NkxxkN: `.origin` includes the port so that works as expected.

Yes good catch it's safer to normalise first and compare.

rishabhpoddar (approved):

PR #175, opened 2021-02-08
Routing refactoring to allow for different matching methods

NkxxkN:
Refactor component routing to allow for different ways of matching with recipe id.

We are associating a `matches` method `() => boolean` to each features. That way, each feature can have its own way to determine wether it should match.

The most common case is using the recipeId which is present in the query params:
https://github.com/supertokens/supertokens-auth-react/compare/0.6...0.6-routing-refactoring?expand=1#diff-52a19574d211f6f85040c50c182de96da31697075d14d68dab12fd95893a32c2R226-R231

But with thirdparty callback routes `auth/callback/github?state={nonce}&code={code}`, there is no `rid` provided in the route. We will define a similar method:

```
export function matchRecipeIdUsingState(recipeId: string): () => boolean {
    return () => {
        const state = getOAuthState();
        if (state === undefined) {
            return false;
        }
        if (state.rid === recipeId) {
            return true;
        }
        return false;
    };
}
```

which will be used for that feature only.


NkxxkN:
Netlify not working because I haven't pushed the `build` for ease of review. 
Nevermind, not needed for that one.

rishabhpoddar:
Why is this function per feature and not per recipe? I think per recipe is better cause there can be features that have nothing to do with routing (like if we make session a feature inside emailpassword).

NkxxkN:
It's not really per feature but per route. A route being responsible of displaying a feature in most cases.
We can't make it by recipe because for example for third party recipe:
 - `/auth?rid=thirdparty` will use query param matching.
 - `/auth/callback/github` will use state param in session storage matching.

PR #179, opened 2021-02-08
third party recipe

NkxxkN:
- Using `0.6-routing-refactoring` as a base as it is has a dependency on it.

Implementation for ThirdParty recipe.


lib/ts/recipe/thirdparty/thirdparty.ts
```diff
@@ -0,0 +1,151 @@
...
+
+        // Add callback route for each provider.
+        this.config.signInAndUpFeature.providers.forEach(provider => {
+            const normalisedFullPath = this.appInfo.websiteBasePath.appendPath(
+                new NormalisedURLPath(`/callback/${provider.id}`)
+            );
+            features[normalisedFullPath.getAsStringDangerous()] = {
+                component: SignInAndUpCallback,
+                rid: this.recipeId,
+                matches: matchRecipeIdUsingState(this.recipeId)
```
NkxxkN, line 86: Match recipe id using state instead of using usual method with query params.

Issue #195, opened 2021-02-17
ThirdPartyEmailPassword recipe


NkxxkN:
## Configs

```js
ThirdPartyEmailPassword.init({

   getRedirectionURL?: (context: {
          action: "SUCCESS";
          isNewUser: boolean;
          redirectToPath?: string;
      }
    | {
          action: "SIGN_IN_AND_UP" | "VERIFY_EMAIL" | "RESET_PASSWORD";
      } | {
       action: "GET_REDIRECT_URL",
       provider: {id: string; name: string}
     }) => Promise<string | undefined>;
     
   preAPIHook?: (context: {
      action: "VERIFY_EMAIL" | "SEND_VERIFY_EMAIL" | "IS_EMAIL_VERIFIED" | "SIGN_OUT" | "SIGN_IN" | " "GET_AUTHORISATION_URL" | "SIGN_UP" | "SEND_RESET_PASSWORD_EMAIL" | "SUBMIT_NEW_PASSWORD";
      requestInit: RequestInit;
   }) => Promise<RequestInit>;
   
   onHandleEvent?: (context:{
          action: "SESSION_ALREADY_EXISTS" | "VERIFY_EMAIL_SENT" | "EMAIL_VERIFIED_SUCCESSFUL" | "RESET_PASSWORD_EMAIL_SENT" | "PASSWORD_RESET_SUCCESSFUL";
      }
    | {
          action: "SUCCESS";
          isNewUser: boolean;
          user: { id: string; email: string };
      }) => Promise<RequestInit>;
  
   palette?: Record<string, string>;
   useShadowDom?: boolean;
   
   disableEmailPassword?: boolean;

   signInAndUpFeature?: {
      disableDefaultImplementation?: boolean;
      defaultToSignUp?: boolean;
      signUpFrom?: {
         privacyPolicyLink?: string;
         termsOfServiceLink?: string
         style?: Record<string, CSSObject>,
         formFields?: {
               id: string;
               label: string;
               placeholder?: string;
               validate?: (value: any) => Promise<string | undefined>;
               optional?: boolean;
         }[]
      },
      signInForm?: {
        style?: Record<string, CSSObject>;
        formFields?: {
             id: string;
             label: string;
             placeholder?: string;
         }[]
      },
      providers?: Provider[]
   },
   resetPasswordUsingTokenFeature?: {
      disableDefaultImplementation?: boolean;
      
      submitNewPasswordForm?: {
[...]

PR #197, opened 2021-02-18
ThirdPartyEmailPassword recipe

NkxxkN:
Implementation for #195


NkxxkN:
🍾 🍾 🍾 🍾 🍾 

https://deploy-preview-197--xenodochial-hamilton-727320.netlify.app/auth?rid=thirdpartyemailpassword&redirectToPath=%2F

rishabhpoddar:
This looks great. But a few points:
- I don't like how there is such less diff between the sign in and sign up UI. It's very simple for an end user to be confused about wanting to sign in, but signin up instead (or vice versa).
- Perhaps an "OR" phrase would be good to separate social login and email password in the UI?

NkxxkN:
>A user should be able to give 0 thirdparty providers. Is that possible?
> A user should be able to disable email password login. Is that possible?

No and no. I had not thought about that. I'll fix it.

> Perhaps an "OR" phrase would be good to separate social login and email password in the UI? 

Just noticed it was in Mufassir's original designs. I'll add.


> I don't like how there is such less diff between the sign in and sign up UI. It's very simple for an end user to be confused about wanting to sign in, but signin up instead (or vice versa).

It's a problem I feel we have already in emailpassword recipe. I've found myself many times getting the field error "This email already exists, please sign in" while I was trying to sign in. What do you suggest?

> Can a user embed the login / reset password etc forms in their own UI?

Yes it works the same way. There are `SignInAndUp` `EmailVerification` and a `ResetPasswordUsingToken` feature components that can be added anywhere in an existing page.

> If this recipe is inside another recipe in the future, will that recipe's ID be carried forward to this recipe and it's child recipes?

The design pattern used here is that we only have one recipe which is responsible for configs & hooks, and routing.
The feature components have all evolved to get their instance recipe using the `recipeId` that is given to them as props.
That means that features can be embedded in each other as long as we make sure to propagate their recipe IDs.
That means any features from `ThirdPartyEmailPassword` can be used within a higher level feature in the future. The theme inside would have to change though.

> Are the various functions exposed from this recipe a union of the functions exposed from the underlying recipes?

Recipe: 
   - Configs: The `ThirdPartyEmailPassword` recipe implements a union of EmailPassword & ThirdParty configs.
   - Routing: Defined on the `ThirdPartyEmailPassword`. This is the only place where I had to duplicate some code from EmailPassword & ThirdParty but this is totally fine.
    - Hooks: Hooks context is a union between ThirdParty and EmailPassword contexts.
    For example: `ThirdPartyEmailPasswordGetRedirectionHookContext = ThirdPartyGetRedirectionHookContext | EmailPasswordGetRedirectionHookContext`
 
 That's all there is to a recipe.
 
 Features:
 As describe in #195, `<ThirdPartyEmailPassword.SignInAndUp>` leverages `<EmailPassword.SignInAndUp>` and `<ThirdParty.SignInAndUp>` features, simply changing the theme that is given to them (just the sign in/sign up form, or just the providers button - no titles/decorations around).
 
 Themes:
 I had to increase EmailPassword & ThirdParty theme's granularity.
 For example the `SignIn` component from EmailPassword was decoupled in:
  - `SignInForm`
  - `SignInHeader`
  - `SignInFooter`
  - `SignIn` which embeds the `SignInHeader`, `SignInFooter` & `SignInForm` and places them inside a container.
  
  That way we are able to reuse the theme components in ThirdPartyEmailPassword features.

rishabhpoddar:
> It's a problem I feel we have already in emailpassword recipe. I've found myself many times getting the field error "This email already exists, please sign in" while I was trying to sign in. What do you suggest?

Same issue here.. @mufassirkazi please fix this UX / UI issue

NkxxkN:
In my opinion, the reason we have this design issue is because we default to signup.

The internet industry has default to sign in: 
 - When you are going to a website with the idea of registering, you look for the "Register" / "Sign Up" button/link.
 - When you already have an account, you just automatically fill in the first form that you see with email / password.

mufassirkazi:
Agreeing with Kevin here. Don't know of any other way that anyone solves this. I have a few ideas but they are too technically complex and no point overengineering this.

rishabhpoddar:
I'm okay with changing default to sign in for emailpassword and thirdpartyemailpassword recipe.

mufassirkazi:
Screen Shot 2021-02-19 at 3 23 56 PM

One of the simplest way we can do this is as represented in the image above. So as soon as the person starts typing we show him the green generic message. Will be kinda annoying though. Maybe give the user the option to enable/ disable?

NkxxkN:
Cool :) 
We can do it starting from `0.7` or we can add a patch to `0.X` ( X < 7) versions too. 
That would be kinda breaking change for `0.X`.
What do you think?

rishabhpoddar:
> That would be kinda breaking change for 0.X. What do you think?

It's OK if it's a breaking change for 0.7 since we are incrementing the first non zero number. So we should make it sign in default from >= 0.7

rishabhpoddar:
@mufassirkazi I did not like this friendly reminder prompt.

NkxxkN:
Merging this.
Opening a new issue for fixing the remarks on this PR => https://github.com/supertokens/supertokens-auth-react/issues/198

PR #200, opened 2021-02-22
add linter rules to make sure jsx is the first import

NkxxkN:
Make sure that emotion `jsx` is imported as the first import when used in a file.

Fix issue with embedding components in a page.

### DIFF
--- a/eslint/index.js
+++ b/eslint/index.js
@@ -1,5 +1,23 @@
 module.exports = {
     rules: {
+
+        "jsx-first-import": {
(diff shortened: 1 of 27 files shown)


NkxxkN:
From [emotion doc](https://emotion.sh/docs/typescript)

image

My understanding is that if you import it after you imported a react element, the `css` property is not going to be recognised and transformed properly. 
That usually leads to 

```
<div css=[object Object] />
```
and no styling is applied.

Issue #210, opened 2021-03-02
Support to overwrite query params from preAPIHooks


NkxxkN:
We currently allow users to overwrite the following from requestInit:
 - body,
 - headers,
 - method
 - integrity
 - cache "default" | "force-cache" | "no-cache" | "no-store" | "only-if-cached" | "reload";
 - keepalive
 - mode ("same-origin" | "cors" | "navigate" | "no-cors")
 - redirect
 - referrer
 - referrerPolicy "" | "no-referrer" | "no-referrer-when-downgrade" | "origin" | "origin-when-cross-origin" | "same-origin" | "strict-origin" | "strict-origin-when-cross-origin" | "unsafe-url";
 - signal
 - window
 
 But `RequestInit` type does not contain `url` and `queryParams`. In order to allow this to be changed we need to expose the `url: RequestInfo` with `requestInfo: Request | string` (Defined by typescript directly).
 
 So modifying the preAPIHook type from:
 
`(context: {requestInit: RequestInit; action: string}) => Promise<RequestInit>` to `(context: {requestInit: RequestInit; action: string}) => Promise<Request | {RequestInit, url}>`.

NkxxkN:
Will be released in 0.7

PR #211, opened 2021-03-03
0.7 preapihook

NkxxkN:
Backward compatible update to preapihook to support updating the URL


NkxxkN:
fixes https://github.com/supertokens/supertokens-auth-react/issues/210

PR #214, opened 2021-03-04
0.7 bug fix path normalisation

NkxxkN:


lib/ts/normalisedURLPath.ts
```diff
@@ -84,7 +83,7 @@ export function normaliseURLPathOrThrowError(input: string): string {
 
 function domainGiven(input: string): boolean {
     // If no dot, return false.
-    if (input.indexOf(".") === -1) {
+    if (input.indexOf(".") === -1 || input.startsWith("/")) {
```
NkxxkN, line 86: our first idea with `input.split("/").filter(...)` was not working because of https://github.com/supertokens/supertokens-auth-react/pull/214/files#diff-8553be74947a88082b1e6e718d428c241c1b3a62c13fab63245a8c1a6b2cb541R91

NkxxkN: Checking for `/` as the first character is actually way enough.

rishabhpoddar: But what if the user gives the path as ".netlify/somethingelse/"

NkxxkN: it clashes with `.example` which we expect to be `.example` (for session and cookies).

Issue #248, opened 2021-05-12
Forms lose style if peer dependency is updated


gusfune:
This is a follow-up on #246 which I have managed to properly replicate after a few tests. It seems a newer version of @emotion (probably @emotion/react@11.4.0) has a breaking change that avoids the rendering of super tokens components. This also conflicts with any other plugin that is using @emotion on a newer version, causing the following console warning:

`You are loading @emotion/react when it is already loaded. Running multiple instances may cause problems. This can happen if multiple versions are used, or if multiple builds of the same version are used.`

Using:
- supertokens-auth-react: 0.12.1
- supertokens-node: 5.0.1
- supertokens-website: 7.0.1   

<img width="376" alt="Screen Shot 2021-05-10 at 17 39 30" src="https://user-images.githubusercontent.com/1147240/117687619-42568f80-b1b8-11eb-99e8-7c1463e55ef3.png">

rishabhpoddar:
Thanks for the issue. Do you think simply updating the emotion version that we use is a good fix? Cause what if this same issue happens in the future with a newer version of emotion? Any ideas on how we can permanently fix this problem?

rishabhpoddar:
In a test app, I installed the latest version of `@hookform/devtools` and `react-hook-form`, and added the `DevTool` component like so:
```js

render() {
const { register, control, handleSubmit } = useForm({
        mode: "onChange",
});

return (
        <>
           <Router>
             ...
           </Router>
           <DevTool control={control} />
        </>
    );
}

```

I do get the warning you mentioned:
```
You are loading @emotion/react when it is already loaded. Running multiple instances may cause problems. This can happen if multiple versions are used, or if multiple builds of the same version are used.
```

However, the form still loads correctly for me:
<img width="1280" alt="Screenshot 2021-05-12 at 21 15 37" src="https://user-images.githubusercontent.com/2976287/118005035-a53a5900-b367-11eb-8614-3c01071c3efb.png">

What am I missing so that I can replicate this issue?

gusfune:
We are using yarn `2.x` instead of npm or yarn classic, and we use deduplicate as well to keep only the most recent version of a package. That might be influencing the issue. I did one more test and if I try to use yarn in pnp mode, the same issue happens no matter if I updated other packages using the same dependencies.

rishabhpoddar:
[...]

PR #249, opened 2021-05-13
Use "^" for @emotion/* packages in package.json

heracek:
Hi,
I'm happy user of `supertokens-auth-react`. I'm using it on site build with [Chakra-UI](https://chakra-ui.com/) which is based on `emotion`.

## Problem

I can't upgrade `@emotion/react` to new versions, because `supertokens-auth-react` has "exact" dependancy on one version of emotion.

If I try to upgrade to new version, it will install two versions of `emotion` (one for my project one for `supertokens-auth-react`) and this is problematic because `emotion` will start complaining about incompatible contexts. And even if it would work it would bundle two versions of `emotion` to my app which would result to increased JS bundle size.

## Solution

Switch to `"^11.2.0"` version for all `@emotion/*` dependencies in `supertokens-auth-react`.

This allows to upgrade `emotion`, hoist it to the root `node_modules` folder and share the same package for my project and 
`supertokens-auth-react`.

## Additional Points

- move all `"dependencies"` to `^` versions (`chroma-js`, and `react-shadow`) (so if projects uses those packages too it will not bundle 2 versions)
- add missing return type for `redirectToLogin` (`npm run lint` was complaining about it and it was easy fix)
- moved `"@emotion/babel-plugin"` to `devDependencies` (it [should be there](https://github.com/emotion-js/emotion/tree/main/packages/babel-plugin#installation))
- upgraded `babel-plugin-inline-react-svg` (`npm audit` was complaining about "high vulnerability" issue)


rishabhpoddar:
Thanks for this PR @heracek. Do you think this PR would solve this issue as well: https://github.com/supertokens/supertokens-auth-react/issues/248?

rishabhpoddar:
Also, if we are upgrading emotion, perhaps we should upgrade to the latest version (to `^11.4.0`). However, that would still be a different version to that used by Chakra-UI... Any ideas on how this issue can be resolved permanently?

heracek:
Thanks for quick merge!


I think it will help with the issue. But I'm not using Yarn v2, so I cannot say for sure that it will fix that issue.

I think that in some cases it will still be possible that two versions of `emotion` will be installed (for example when you upgrade packages), but running `yarn upgrade --deep` should fix the issue.

To completely fix the issue I think it will be necessary to move all `emotion` dependencies to `"peerDependencies"`. But this would mean that users of `supertokens-auth-react` would have to run: `yarn install supertokens-auth-react @emotion/react` to make `supertokens-auth-react` work.

rishabhpoddar:
Makes sense.. for now i'll just upgrade the package to use the latest version of emotion with a `^`. If this issue comes up in the future again, then perhaps can move it to peer dependency...

heracek:
> Also, if we are upgrading emotion, perhaps we should upgrade to the latest version (to `^11.4.0`). However, that would still be a different version to that used by Chakra-UI... Any ideas on how this issue can be resolved permanently?

I intentionally left the version to lower `^11.2.0` (vs. latest `^11.4.0`). This will "not force" me to upgrade `emotion` when I'll be using `supertokens-auth-react` (for example I have to stay on older version because of some other package).

I think `supertokens-auth-react` should depend on lowest possible and use `^` to be compatible with wide range of `emotion` packages. If there is issue in some specific `emotion` version that has effect on `supertokens-auth-react`, then you should increase the version number.

rishabhpoddar:
> I think supertokens-auth-react should depend on lowest possible and use ^ to be compatible with wide range of emotion packages.

Agreed! Thanks

rishabhpoddar:
@heracek this change has been release in version 0.13.0.

heracek:
Thanks 🤩

PR #251, opened 2021-05-24
Ability to customise recipe functions

rishabhpoddar:
- [x] Changes
    - [x] Makes types simpler
    - [x] Make emailverification into a recipe
    - [x] Eliminate the need for getRecipeOrThrow in feature and theme components by passing down config
    - [x] Create and use interfaces for all recipes
    - [x] Allow interfaces to be overrided
    - [x] Any better way of doing preApiHook?
    - [x] make httpRequest part of implementation as opposed to part of recipe
    - [x] Remove httpRequest from super recipe class
    - [x] move authRecipeModule > theme into each recipe's theme.
    - [x] Move wrappers components to their own recipe.
    - [x] Move with-ts example in test folder, check for types, and remove all .ts files that are only for importing purposes
    - [x] Override components and style
        - How will the user find the actual implementation of a component on github? - https://github.com/supertokens/supertokens-auth-react/issues/260
    - [x] Do we need GET_REDIRECTION_URL in GetRedirectionURLContext in thirdparty?
    - [x] Do not pass status as a child to thirdpary theme
    - [x] redirectToAuth should not add redirectTo query param by default
    - [x] Event handlers should be at the time of implementation in recipe
    - [x] Test out override in typescript, simple and emailverification override as well.
    - [x] Fix Pre API hook event name for thirdparty sign in up and email password sign in and sign up
- [ ] Tests
    - [x] Manually rendering feature components 
        - [x] One per auth recipe (3 of them)
    - [ ] Make sure react router dom is being used properly
        - This may be hard to test. But one possible idea is to store some variable in memory on test start, and navigate a bunch of pages (sign in, sign out etc...), and then check that the variable is still set at the end of the test. If at anytime, the page has been reloaded (cause maybe a bug in not using react-route-dom properly), then the variable will have disappeared and the test would fail..
    - [ ] Email verification override
        - [ ] One per auth recipe (3 of them)
    - [ ] Normal override
        - [ ] all recipes (4 of them) 
    - [ ] Check in prod app if you see a big wall of red error when you navigate to /auth/verify-email?token=TOKEN&mode=REQUIRED.
        - This is manual test, can ignore.
    - [x] Component override tests
    - [x] NormalisationPath change for when the path is "/"
        - No need to explicitly test this.. it's just FYI for me to manually test 
    - [x] Session override, onHandleEvent and preAPIHooks
        - Add preAPIHooks and onHandleEvent callbacks to Session.init (with proper formatting). And check that in end-to-end tests, those are being added to the console log. You will need to change the console logs checking to get them to pass, but when you change them, make sure they make sense (for ex, check that SIGN_OUT event is getting fired at the right time in the sequence of logs)
        - One issue might come up with refresh session related events - they may behave differently on each run (though they should not). In that case, you can omit logging those events in case the `context.action` is refresh (in the `preAPIHooks` and `onHandleEvent`).
    - [ ] Is it fine that prop-type is in devDependency?
        - One time manual test.. please ignore


dulowski-marek:
> Is it fine that prop-type is in devDependency?

If it's not being removed during build step from production code, it should be listed as `dependency` or `peerDependency`.

PR #258, opened 2021-06-08
Detects if withOverride is being using without parent provider

rishabhpoddar:
@dulowski-marek  please feel free to merge into your branch once you have reviewed it

Related to: https://github.com/supertokens/supertokens-auth-react/pull/255

### DIFF
--- a/lib/ts/components/componentOverride/componentOverrideContext.ts
+++ b/lib/ts/components/componentOverride/componentOverrideContext.ts
@@ -1,8 +1,10 @@
 import React from "react";
 import { ComponentOverride } from "./componentOverride";
 
-type ContextType<T> = {
-    [K in keyof T]?: ComponentOverride<any>;
-};
+type ContextType<T> =
+    | {
+          [K in keyof T]?: ComponentOverride<any>;
+      }
+    | "IS_DEFAULT";
 
-export const ComponentOverrideContext = React.createContext<ContextType<any>>({});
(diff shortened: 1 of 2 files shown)


lib/ts/components/componentOverride/componentOverrideContext.ts
```diff
@@ -1,8 +1,10 @@
...
 import { ComponentOverride } from "./componentOverride";
 
-type ContextType<T> = {
-    [K in keyof T]?: ComponentOverride<any>;
-};
+type ContextType<T> =
+    | {
+          [K in keyof T]?: ComponentOverride<any>;
+      }
+    | "IS_DEFAULT";
```
dulowski-marek, line 8: Why use `IS_DEFAULT` string instead of `undefined` or `null` for instance?

dulowski-marek (commented):
Just curious about the reason

lib/ts/components/componentOverride/componentOverrideContext.ts
rishabhpoddar: Just to be more explicit.. Could have used `undefined` or `null` as well.. But they we would not be able to differentiate if the user has provided `undefined` as opposed to it being our mistake where we forgot to add a parent provider.

dulowski-marek: Right, the mistake might also be related to passing invalid value in normalize config as well. Let's go this way.

dulowski-marek (approved):

Issue #257, opened 2021-06-08
Expose all recipe interface functions for each recipe


rishabhpoddar:
- If a recipe interface has 5 functions, we should expose all of them via the recipe's index.ts file so that the user can call them manually if they want.
- Also change [this guide](https://github.com/supertokens/supertokens-core/wiki/Development-guideline-for-the-backend-and-frontend-recipes) to say that this needs to be done for any new recipe
- Change to docs as well.

rishabhpoddar:
Closing in favour of https://github.com/supertokens/for-zenhub/issues/28

PR #261, opened 2021-06-10
adds "OPTIONAL" mode for email validation

aldeed:
Related to https://github.com/supertokens/supertokens-core/issues/159

This does not fully implement the linked issue, but it is useful in that it allows optional validation to be done. Without this there is no way to do it because "OFF" does not create the necessary routes while "REQUIRED" adds an undesired redirect.

This PR adds "OPTIONAL", which does create the necessary routes but does not redirect, and `isEmailVerificationRequired` continues to return `false`.

The way I am using this is in a NextJS app, with React browser code similar to this:

```jsx
<Button onClick={handleVerifyClick}>Verify Email Address</Button>
```

```ts
const handleVerifyClick = async (): Promise<void> => {
    if (isEmailVerified) return

    const response = await fetch(`${appInfo.apiBasePath}user/email/verify/token`, {
      method: 'POST',
      headers: {
        'Content-Type': 'application/json',
        'fdi-version': '1.8',
        rid: 'emailpassword'
      }
    })
    const responseObj = await response.json()
    if (responseObj.status === 'OK') {
      openEmailVerificationSentSnackbar('We sent you an email. Click the button in the email to verify your address.')
    }
  }
```

I think it would be good to also export a `sendVerifyEmail` API util function in browser code, similar to how `isEmailVerified` is exported, so that I don't have to hack the fetch call. But I'm a little lost about how to properly make this change.

I would also need some guidance about where to update or add tests for this change.

The issue mentions having built-in UI for optional validation. I guess this would be similar to what exists but requiring a button click to send the email. I actually would prefer if "REQUIRED" UI did not automatically send the email on mount. It tends to get stuck in a redirect cycle sometimes, which causes it to send many emails in a row. If "REQUIRED" and "OPTIONAL" both require you to click a button first, it would solve both problems. I could simply manually redirect to the router, or render `EmailVerification` component anywhere, without worrying about either one automatically sending emails.

Anyway, just wanted to provide a starting point for this feature.

### DIFF
--- a/lib/build/recipe/emailverification/types.d.ts
+++ b/lib/build/recipe/emailverification/types.d.ts
@@ -4,7 +4,7 @@ import { ComponentOverride } from "../../components/componentOverride/componentO
 import { SendVerifyEmail } from "./components/themes/emailVerification/sendVerifyEmail";
 import { VerifyEmailLinkClicked } from "./components/themes/emailVerification/verifyEmailLinkClicked";
 export declare type UserInputForAuthRecipeModule = {
-    mode?: "OFF" | "REQUIRED";
+    mode?: "OFF" | "REQUIRED" | "OPTIONAL";
     disableDefaultImplementation?: boolean;
     sendVerifyEmailScreen?: FeatureBaseConfig;
     verifyEmailLinkClickedScreen?: FeatureBaseConfig;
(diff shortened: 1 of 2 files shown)


rishabhpoddar:
Thanks @aldeed for this PR. I'll come back to this feature soon enough to complete it :)
About the missing function being exposed, we have an open issue for this: https://github.com/supertokens/supertokens-auth-react/issues/257

rishabhpoddar:
@aldeed can you please elaborate on "It tends to get stuck in a redirect cycle sometimes"? There shouldn't be any redirect cycles as such..

aldeed:
@rishabhpoddar When I tried "REQUIRED" mode or put the validation component on a page, it would continually refresh the page and send an email each time it refreshed. I'm guessing it's something weird about my config or the way I'm doing things, but I didn't have time to dig deeply into it. I do have a custom `getRedirectionUrl` but when I commented that out it still happened.

rishabhpoddar:
@aldeed thanks for the info. When you say "put the validation component on a page", are you referring to Step 2 in [this page](https://supertokens.io/docs/thirdpartyemailpassword/common-customizations/email-verification/embed-in-page)?

aldeed:
Yes

rishabhpoddar:
Hey! This PR is quite outdated now and this mode would be easily do-able with our new session grant feature (which we are working on now). So closing this PR.

PR #265, opened 2021-06-14
Test embeding component behaviour

dulowski-marek:
### Related
- https://github.com/supertokens/supertokens-auth-react/pull/251

### Scope
Test the behaviour described here https://supertokens.io/docs/thirdpartyemailpassword/common-customizations/signin-form/embed-in-page

### Remaining
- [x] `/auth/reset-password`
- [x] `/auth/verify-email`


rishabhpoddar:
`disableDefaultImplementation` can be given to email verificaiton and reset password features too.

dulowski-marek:
> disableDefaultImplementation can be given to email verificaiton and reset password features too.

Yes, I'm going to push another few commits.

test/end-to-end/pages/AuthPage.js
```diff
@@ -0,0 +1,30 @@
...
+        await Promise.all([page.goto(url), page.waitForNavigation({ waitUntil: "networkidle0" })]);
+
+        return new AuthPage(page);
+    }
+
+    constructor(page) {
+        this.page = page;
+    }
+
+    async isFeatureMounted() {
```
rishabhpoddar, line 19: Have you tested that this function returns false if the feature is not there? One way to test this is to intentionally introduce a bug a feature component where it doesn't render the theme, and then see if this returns false.

examples/for-tests/src/App.js
```diff
@@ -403,14 +409,14 @@ function getThirdPartyEmailPasswordConfigs() {
...
         },
-
         useShadowDom,
         palette: theme.colors,
         emailVerificationFeature: {
             mode: emailVerificationMode,
         },
         resetPasswordUsingTokenFeature: {},
         signInAndUpFeature: {
+            disableDefaultImplementation,
```
rishabhpoddar, line 419: If this is set to `false`, you need to render the feature component manually in a route - I don't see that change in the for-tests application.. how are the tests passing?

rishabhpoddar (changes requested):

examples/for-tests/src/App.js
dulowski-marek: You're right. The behaviour I'm testing is:
- if you set `disableDefaultImplementation: true` on a feature
- then supertokens feature UI shouldn't render on `/default-path-for-feature` route

There are two different behaviours that this usecase is composed of:
1. Not rendering supertokens UI on `/somepath` - this is tested, as this is the primary concern for using this feature by users
2. Rendering some feature component outside - this is not tested and I would leave that for a unit test, since the behaviour is no different from rendering it normally (without `disableDefaultImplementation`). Unless it interacts with the previous behaviour in some special way that I didn't discover.

Let me know please for the second case if there's something I missed.

rishabhpoddar: - I don't think I understand point number 1. What is `/somepath` supposed to show?
- So are you simply testing that the feature does not show when you set `disableDefaultImplementation: true`?

test/end-to-end/pages/AuthPage.js
dulowski-marek: It relies on application being rendered on a route - it will return `false` when the supertokens UI is not mounted.

I can improve the selector, it will be different for each route, but will become more independent on how the top-level components are rendered.

dulowski-marek:
Marked as draft, because `/auth/verify-email` tests timeout on navigation.

examples/for-tests/src/App.js
dulowski-marek: > - I don't think I understand point number 1. What is /somepath supposed to show?

It's just for illustrative purpose - the real paths we don't render on are `/auth`, `/auth/verify-email`, `/auth/reset-password`. I could have been clearer, sorry.

> - So are you simply testing that the feature does not show when you set disableDefaultImplementation: true?

Exactly. To be specific, I'm testing that it shows with `disableDefaultImplementation: false` and then that it doesn't with `disableDefaultImplementation: true`.

test/end-to-end/pages/AuthPage.js
dulowski-marek: I'm going to extract that to an issue, as at this moment it is sufficient to check if the UI was mounted. Tracked in https://github.com/supertokens/supertokens-auth-react/issues/266

examples/for-tests/src/App.js
dulowski-marek: I'm going to resolve this, based on our discussion on the call.

Issue #241, opened 2021-05-02
Use SessionAuth with requireAuth false in all components controlled by us


rishabhpoddar:
This will remove the jittering effect with login pages are loaded

Issue #228, opened 2021-04-14
Enhance useSessionContext & SessionAuth


rishabhpoddar:
- Update to jwt payload
- Update to doesSessionExist

This will then allow us to create a wrapper for the whole app and the user can then access this context from anywhere easily (Keep in mind this issue though: https://github.com/supertokens/supertokens-auth-react/issues/263)!

------------
The tricky part of this task:
- initially, we may think that we can simply fire an event in the supertokens-website repo in the relevant places, and listen for them in the SessionAuth wrapper. This may work for our default implementation, but what if someone decides to override how the session logic works entirely? Then those events will not get fired.
- We can ask the user to fire those events for us in that case, but is that the best way of doing it?

rishabhpoddar:
A few points to consider while solving this issue:
- Right now, if the user calls sign out, they need to navigate to the auth page themselves. However, with this change, that's not true anymore (unless `requireAuth` prop is `false`).
- If a session has expired (`UNAUTHORISED` event fired by supertokens-website), then the user will be immediately navigated to the auth page - I think this is OK for now, but later on, we may want to show a popup saying something like session has expired.

Issue #271, opened 2021-06-17
Make override components feature more flexible


rishabhpoddar:
- Right now, we have to pick specific components and make them overridable
- Ideally, we want to allow the user to be able to override any component in the component tree
- We have to be careful to account for overriding a specific instance of a reusable component

rishabhpoddar:
Checkout how `swizzle` works for Docusaurus v2.. It allows you to "eject" any component in their component tree into a specific folder location and the user can change that react component however they like.

This seems to be a great method of achieving what we wanted to in the first place!

rishabhpoddar:
Closing due to inactivity.

PR #274, opened 2021-06-22
Implement enhanced session context update API

dulowski-marek:
### Related
- https://github.com/supertokens/supertokens-auth-react/issues/228
- https://github.com/supertokens/supertokens-auth-react/issues/241

The draft illustrates the interface API for SessionContext updates mechanism.

### Scope & Reasoning
1. Introduce `Authentication` object within `supertokens-website` representing auth state at given point in time.
```ts
interface Authentication {
    isAuthenticated(): boolean;
    getUserid(): string; // throws if not authenticated
    getJwtPayload(): any; // throws if not authenticated
}
```
This may seem redundant, because `SessionContext` provides similar data structure. The goal though is to represent this concept of auth state in `supertokens-website` and then just be able to map it wherever it is required.

2. Listen to `Authentication` objects whenever the state changes.
This can be done using a simplified observer pattern. I'm hesitant to use `onHandleEvent`, since the API is public and it would make it more difficult to change, if we want to refactor for NextJS usage, for example.

3. Map `Authentication` to `SessionContext` to preserve compatibility with existing API.
This separates user-oriented APIs from those provided by library by default. In case we need to change how recipes behave, we'll still have a clear and concise API to subscribe to auth state, that cannot be overridden.

## Pending TODOs:
- [x] How does NextJS SSR impact the implementation?
- [x] Does this PR affect components override?
- [ ] supertokens-website test for reading userId right after create new session event is fired.
- [x] More tests for this PR?
    - [x] If the JWT payload has changed (due to refreshing), then it is reflected in the context correctly. 
- [ ] [This test](https://github.com/supertokens/supertokens-auth-react/blob/3a627e08fc96f54412e33d1252c0f0da127f2ce2/test/end-to-end/emailverification.test.js#L363) has been commented out because it doesn't actually stop the server in [this line](https://github.com/supertokens/supertokens-auth-react/blob/3a627e08fc96f54412e33d1252c0f0da127f2ce2/test/end-to-end/emailverification.test.js#L372) - cause there is no `/stop POST` API in the server.. Perhaps another way to make `isEmailVerified` function throw a 500 error to test this out. Plus, post [this line](https://github.com/supertokens/supertokens-auth-react/blob/3a627e08fc96f54412e33d1252c0f0da127f2ce2/test/end-to-end/emailverification.test.js#L382), we should check that the contents of the screen have been rendered.


lib/ts/recipe/session/recipe.ts
```diff
@@ -19,18 +19,27 @@
...
+    constructor(config: ConfigType) {
         super(config);
 
+        if (config.onHandleEvent !== undefined) {
+            this.addEventListener(config.onHandleEvent);
+        }
+
         sessionSdk.init({
             ...config,
+            onHandleEvent: this.onHandleEvent,
```
rishabhpoddar, line 42: We need to call the user's `onHandleEvent` as well?

lib/ts/recipe/session/types.ts
```diff
@@ -15,6 +15,10 @@
 
 import { RecipeInterface } from "supertokens-website";
 
+export type OnHandleEventContext = {
+    action: "SIGN_OUT" | "REFRESH_SESSION" | "UNAUTHORISED";
```
rishabhpoddar, line 19: We also need to add one more event for `NEW_SESSION_CREATED`?

rishabhpoddar (changes requested):

lib/ts/recipe/session/recipe.ts
dulowski-marek: Not exactly. The handler itself will be called by this simplified listener/observer code https://github.com/supertokens/supertokens-auth-react/pull/274/files#diff-dd812fd78904a37e7eb4dc6156f8e62e23cf966041cfc5d89120378b0dd5e9d6R101

We add the user's event handler here https://github.com/supertokens/supertokens-auth-react/pull/274/files/1fec5f5621ab9d9ccdfedf6c0c65055d3fe48d1f#diff-dd812fd78904a37e7eb4dc6156f8e62e23cf966041cfc5d89120378b0dd5e9d6R36

So `Session` RecipeModule will call all the listeners that we have in `eventListeners` set. This includes user event listener and `SessionAuth` event listener.

dulowski-marek: So user's `onHandleEvent` will be called, but not by decorating the function.

lib/ts/recipe/session/sessionAuth.tsx
```diff
@@ -16,97 +16,82 @@
...
         }
+    };
+
+    // Read the current state
+    useEffect(() => {
+        updateSessionContext();
+    }, []);
+
+    // Setup listener
+    useEffect(() => session.addEventListener(onHandleEvent), []);
```
rishabhpoddar, line 80: Do we need to remove eventListener when this component is unmounted?

rishabhpoddar (changes requested):

lib/ts/recipe/session/recipe.ts
rishabhpoddar: Sounds good. Sorry, missed that line of code whilst seeing it...

lib/ts/recipe/session/sessionAuth.tsx
dulowski-marek: It will be removed, because `addEventListener` returns an unsubscribe function. `useEffect` with empty deps array will cleanup when the component is unmounted.

dulowski-marek: Just as you advised, I will document this function better.

lib/ts/recipe/session/recipe.ts
```diff
@@ -91,13 +91,16 @@ export default class Session extends RecipeModule<unknown, unknown, unknown, any
         return sessionSdk.attemptRefreshingSession();
     };
 
-    addEventListener(listener: (ctx: OnHandleEventContext) => void): () => void {
+    /**
+     * @returns Function to remove event listener
+     */
+    addEventListener(listener: (ctx: RecipeEvent) => void): () => void {
```
rishabhpoddar, line 97: Use arrow functions, other you have to bind `addEventListener`

lib/ts/recipe/session/sessionAuth.tsx
```diff
@@ -31,67 +32,77 @@ type Props = {
...
             case "REFRESH_SESSION":
+                return setContext({
+                    doesSessionExist: true,
+                    userId: context === undefined ? "" : context.userId,
+                    jwtPayload: await getJWTPayloadSecurely(),
+                });
             case "SIGN_OUT":
             case "UNAUTHORISED":
-                return updateSessionContext();
+                return setContext({
```
rishabhpoddar, line 75: Setting this here is actually a breaking change since previously, users had to redirect to auth page after sign out or session expiry manually.

lib/ts/recipe/session/sessionAuth.tsx
```diff
@@ -31,67 +32,77 @@ type Props = {
...
     // If the context is undefined, we are still waiting to know whether session exists.
-    if (sessionContext === undefined) {
+    if (context === undefined) {
         return null;
     }
 
     // If the session doesn't exist and we require auth, redirect to login
-    if (sessionContext.doesSessionExist === false && props.requireAuth === true) {
+    if (context.doesSessionExist === false && props.requireAuth === true) {
         props.redirectToLogin();
```
rishabhpoddar, line 100: This should be called in `onHandleEvent` or in `setInitialContext`

rishabhpoddar (changes requested):

lib/ts/recipe/session/sessionAuth.tsx
dulowski-marek: I think it makes sense to make this change and bump the version appropriately.

If we have those automatic context updates, then if a user is using `SessionAuth` (indirectly, through the library) the state is consistent, because we do not allow `SessionAuth` to render its children if there's no session, which is one concern of this component.

What do you think?

lib/ts/recipe/session/sessionAuth.tsx
dulowski-marek: > This should be called in `onHandleEvent`...

Again, I would limit the responsibility of `onHandleEvent` function to just setting the state. Then, take actions based on the state. This makes it a lot clearer, as this is a common pattern :)

> ...or in `setInitialContext`
I disagree. `setInitialContext` should only care about setting context. We may do it in `useEffect` hook next to `setInitialContext`, as a separate function.

This would make it only react to the current state, when the component is rendered, which I assume is not what we want.

Summarizing, I would leave that in render block.

rishabhpoddar: Agreed. However, I meant to say that shouldn't a side effect like `props.redirectToLogin();` be called in something like `useEffect` or something other similar function? Cause then it would be like calling `props.redirectToLogin();` instead the `render()` function? 

I am not too familiar with the exact semantics of functional components... so if you think this is fine, then cool!

rishabhpoddar: So I think functional components are called multiple times (each render). This means, if this function is called twice very fast when `context.doesSessionExist === false && props.requireAuth === true` is true, you will be calling the user's `redirectToLogin` function twice.. (See [this blog](https://dmitripavlutin.com/react-useeffect-explanation/) - "The component rendering and side-effect logic are independent. So it would be a mistake to perform side-effects directly in the body of the component.")

Since calling `props.redirectToLogin();` depends on `context` and `props`, maybe you could:
```
useEffect(() => {
    // Runs ONCE after initial rendering
    // and after every rendering ONLY IF `prop` or `state` changes
    if (sessionContext.doesSessionExist === false && props.requireAuth === true) {
       props.redirectToLogin();
    }
  }, [prop, context]);
```

rishabhpoddar:
@dulowski-marek answering one of the open questions in this PR:

> How does NextJS SSR impact the implementation?

It doesn't, cause we tell NextJS to not render auth wrapper components on the backend.

rishabhpoddar:
I think that we should add a new prop to all auth wrappers -> `onSessionExpired?: () => void` which we will call (if present) when the `UNAUTHORISED` event is fired. If it's not present, we will continue to do what we are doing now - which is redirecting the user.

This will allow the end user to show a popup or something that the session has expired and then redirect them to the login page on after the use dismisses their popup.

dulowski-marek:
> Does this PR affect components override?

### For the current state of components override (before https://github.com/supertokens/supertokens-auth-react/issues/271 is resolved):
1. **Behaviour**: The isolated override behaviour didn't break, as the unit tests that I prepared are passing. The only way it could interact is due to changes you made to feature components, although after reviewing it I doubt it will cause problems. This however should be a subject to integration tests (which we don't have in this repo). 

2. **Component tree**: The other thing influenced is component tree, as you added components. This will not be a problem though, as component names and available components stay the same.

3. **Context**: I don't see any scenario in which it would interact with `ComponentOverride` context.

4. **Overridden components that rely on `SessionContext`**: This is more tricky, but is handled with context preservation on `UNAUTHORISED` which you provided.

### For the future implementation (after https://github.com/supertokens/supertokens-auth-react/issues/271 is resolved):
1. **Overridden components that rely on `SessionContext`**: This will be handled with similar mechanism as currently (preserving context).

### Summary
I don't think it will interact with component override currently. For the future, since we are clearly separating the API for a consumer from internal workings, I think we'll be able to make internal adjustments so that we are not constrained when implementing https://github.com/supertokens/supertokens-auth-react/issues/271.

dulowski-marek:
> I think that we should add a new prop to all auth wrappers -> `onSessionExpired?: () => void` which we will call (if present) when the `UNAUTHORISED` event is fired. If it's not present, we will continue to do what we are doing now - which is redirecting the user.
> 
> This will allow the end user to show a popup or something that the session has expired and then redirect them to the login page on after the use dismisses their popup.

I like the idea. The user should be able to act on event and not rely on default behaviour. 
- How should that be passed in config?
- How about the current redirection happening on `UNAUTHORISED` event? Are we going to make it default and just allow to override with `onSessionExpired`?

rishabhpoddar:
> How should that be passed in config?

It will be an optional prop to the auth wrapper components

> How about the current redirection happening on UNAUTHORISED event? Are we going to make it default and just allow to override with onSessionExpired?

Yes.

lib/ts/recipe/session/sessionAuth.tsx
```diff
@@ -54,33 +54,51 @@ const SessionAuth: React.FC<Props> = ({ children, ...props }) => {
...
+                        userId: "",
+                        jwtPayload: {},
+                    });
+
+                    return;
+                case "UNAUTHORISED":
+                    // If there's onSessionExpired handler, use it without setting state...
+                    if (onSessionExpired !== undefined) {
+                        onSessionExpired();
+
```
rishabhpoddar, line 86: Please remove empty lines like there :)

rishabhpoddar (changes requested):

rishabhpoddar:
Two issues:
- When using `SessionAuth` wrapper, it makes sense to provide the `redirectToLogin` callback only when `requireAuth` is `true`. Perhaps we can make the input prop type better defined.
- If someone decides to wrap their entire app with `EmailPasswordAuth`, then there can be an infinite redirection loop happening if they visit the email verification screen.

dulowski-marek:
> * If someone decides to wrap their entire app with `EmailPasswordAuth`, then there can be an infinite redirection loop happening if they visit the email verification screen. We can solve this by checking of the current URL path is the same as the email verificaiton URL path, and if it is, we do not redirect?

As discussed - this issue is not directly related to changes in this PR, but rather originates in the chosen approach of using auth wrappers to do two things: provide context and act on context changes. Let's handle it separately to prevent scope creep https://github.com/supertokens/supertokens-auth-react/issues/278

lib/ts/recipe/session/RequireSession.tsx
```diff
@@ -0,0 +1,16 @@
+import React, { useContext } from "react";
+import SessionContext from "./SessionContext";
+
+type Props = {
+    requireSession?: boolean;
+};
+
+export const RequireSession: React.FC<Props> = ({ children, requireSession = true }) => {
```
rishabhpoddar, line 8: - I don't really see a point in having this refactor, especially since you can give `requireSession` as `false`.
- If this component must exist, it should not take a prop and should only show `children` if `doesSessionExist`.

lib/ts/recipe/session/index.ts
```diff
@@ -18,7 +18,7 @@ import { RecipeInterface } from "supertokens-website";
 import SessionAuthWrapper from "./sessionAuth";
 import useSessionContextFunc from "./useSessionContext";
 import { InputType, SessionContextType } from "./types";
-import SessionContext from "./sessionContext";
+import SessionContext from "./SessionContext";
```
rishabhpoddar, line 21: IMPORTANT: I had renamed this file to be `sessionContext`. The capital letter at the start causes issues with Github since github  seems to ignore capital letter starts to the name of the file..

lib/ts/recipe/session/useSessionContext.ts
```diff
@@ -1,6 +1,6 @@
 import React from "react";
 
-import SessionContext from "./sessionContext";
+import SessionContext from "./SessionContext";
```
rishabhpoddar, line 3: Should be small letter.

lib/ts/recipe/session/RequireSession.tsx
dulowski-marek: I extracted this component for higher granularity, so that `SessionAuth` can be composed out of smaller, easier testable components like this. I'm certain that such refactors will speed up future recipe development and limit false reuse.

`requireSession` is there, because it's designed to provide similar behaviour as `SessionAuth` has currently, so that it works with how we handle `requireAuth` currently.

Issue #279, opened 2021-06-30
Override components loose state on rerender


rishabhpoddar:
This is because the `withOverride` components calls the function provided by the user on each render, which effectively results in a new component being returned - making what's currently rendered loose all its state.

PR #282, opened 2021-07-01
Fix override re-renders causing state to be lost

dulowski-marek:
### Related
- https://github.com/supertokens/supertokens-auth-react/issues/279

### Scope
- [x] add test for component preserving state on rerender
- [x] fix the state loss issue when component rerenders


dulowski-marek:
I will have to ask for your input @rishabhpoddar - I wasn't able to replicate the issue, so test implemented in https://github.com/supertokens/supertokens-auth-react/pull/282/files#diff-9cf05339604ca69f84820a372d795596c35097c4ac7db5a839a6168134ff4740R51 is how I understand it. 

This test is passing, so I guess the issue lies elsewhere.

dulowski-marek:
So the previously implemented test was wrong in a way, that it tested `DefaultComponent` state. It also didn't pass, because the parent `OverrideComponent` got rerendered, and thus all components inside (including `DefaultComponent`).

New test was added in https://github.com/supertokens/supertokens-auth-react/pull/282/commits/77df5edef34d12ca6f452a89f78e3c800391192d along with implementation change, which breaks how component overrides are defined.

PR #287, opened 2021-07-06
SessionAuth API Proof of Concept

dulowski-marek:
# This should not be merged

This is an example of my idea how `Session` API could look like.

### DIFF
--- a/lib/ts/recipe/session/poc/authentication.ts
+++ b/lib/ts/recipe/session/poc/authentication.ts
@@ -1,0 +1,4 @@
+export type Authentication<T = any> = Readonly<{
+    userId: string;
+    jwtPayload: T;
+}>;
--- a/lib/ts/recipe/session/poc/authenticationExpired.ts
+++ b/lib/ts/recipe/session/poc/authenticationExpired.ts
@@ -1,0 +1,1 @@
+export class AuthenticationExpired { }
--- a/lib/ts/recipe/session/poc/authenticationSource.ts
+++ b/lib/ts/recipe/session/poc/authenticationSource.ts
@@ -1,0 +1,6 @@
+import { AuthenticationExpired } from "./authenticationExpired";
+import { AuthenticationUpdated } from "./authenticationUpdated";
+
+export interface AuthenticationSource {
+    addListener(listener: (event: AuthenticationExpired | AuthenticationUpdated) => void): () => void;
+}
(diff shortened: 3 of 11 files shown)


rishabhpoddar:
So how would the user use it with all the parts:
- they want to protect a route
- do email verification redirects
- have a popup on session expiry
- provide session context to the children

```js
<Session.Provider>
   <EmailPassword.Require whenNotExists={() => {
       return <h1>I'm rendered when session doesn't exist</h1>
    }}>
        <Session.OnExpired render={({ didSessionExpire, session }) => {...}}>
           <EmailVerificationRequired>
              <MyComponent />
           </EmailVerificationRequired>
        </Session.OnExpired>
   </EmailPassword.Require>
</Session.Provider>
```

We can more `<Session.Provider>` to cover the whole app which reduces one level of nesting here, but this won't work for server side rendered app (like NextJS) since `<Session.Provider>` doesn't support SSR.

dulowski-marek:
> So how would the user use it with all the parts:
> - they want to protect a route

This can be done by just using `Session.Require`:
```tsx
<Router>
    <Session.Provide>
        <Route component={SomeComponent} />
    </Session.Provide>
</Router>

const SomeComponent = () => (
    <Session.Require whenNotExists={/* render placeholder or redirecting component */}>
        <h1>Content</h1>
    </Session.Require>
)
```

> - do email verification redirects

About this one I'm not sure. I'm thinking if `EmailVerification` could be built without relying on session apis, apart from context only to get user info.
> - have a popup on session expiry

`Session.OnExpired` is designed to do just that. It should "freeze" the session at its current state (although this is not implemented fully now)
```tsx
<Session.OnExpired render={({ didSessionExpire, session }) => (
    <div>
        <h1>User id is: {session.userId}</h1>
        {didSessionExpire ? <ExpiredModal /> : null}
    </div>
)} />
```
> - provide session context to the children

This is handled by context provider `Session.Provide`:
```tsx
<Session.Provide>
    <h1>I will access SessionContext</h1>
</Session.Provide>
```

> We can more <Session.Provider> to cover the whole app which reduces one level of nesting here, but this won't work for server side rendered app (like NextJS) since <Session.Provider> doesn't support SSR.

What is the problem you see with supporting SSR?

dulowski-marek:
As discussed, the problem with `Session.Provide` supporting SSR is that `doesSessionExists` reads frontend cookie value (just as other functions do, `getJWTPayloadSecurely`, `getUserId`). It's not feasible to wrap the whole application in `Session.Provide`, since it will prevent rendering the authenticated UI.

This original problem is not solved now either and needs more research.

Issue #290, opened 2021-07-07
Fix behaviour of SessionAuth Wrapper


rishabhpoddar:
## TODO:
- [x] confirm the new behavior
- [x] Testing

![Image from iOS.jpg](https://images.zenhubusercontent.com/60df055a26df4b7a2fb68f0e/8882736c-0f16-4da1-ad17-4cc94781c9bb)

And https://docs.google.com/spreadsheets/d/19QGn_vKAbph9Uvtqi2yOF9pJqWp5Xk9mzia1qb0GVdE/edit#gid=0

rishabhpoddar:
Most of the issues for this have been fixed, except for https://github.com/supertokens/supertokens-auth-react/issues/293 and https://github.com/supertokens/supertokens-auth-react/issues/294.

PR #291, opened 2021-07-07
Session auth wrapper fix

rishabhpoddar:
Related to https://github.com/supertokens/supertokens-auth-react/issues/290


lib/ts/components/superTokensRoute.tsx
```diff
@@ -54,8 +54,10 @@ function SuperTokensRouteWithRecipeId({
     withRouter: WithRouterType;
 }): JSX.Element | null {
     const normalisedPath = new NormalisedURLPath(path);
-    const featureComponentWithRecipeId =
-        SuperTokens.getInstanceOrThrow().getMatchingComponentForRouteAndRecipeId(normalisedPath);
+
+    const featureComponentWithRecipeId = React.useMemo(() => {
```
dulowski-marek, line 58: Two categories of issues with this:
1. Style: everywhere we use hooks like `useMemo(...)`, `useRef(...)` directly importing them as `import { useMemo, useRef } from "react"`. I would also stick to this style here.
2. Best practices
- `useMemo` is a solution to memoize a value of expensive computation, not just storing some value across rerenders (which `useRef` is for).
- `useMemo` shouldn't be relied on for behaviour of the component: https://reactjs.org/docs/hooks-reference.html#usememo

> You may rely on useMemo as a performance optimization, not as a semantic guarantee. In the future, React may choose to “forget” some previously memoized values and recalculate them on next render, e.g. to free memory for offscreen components. Write your code so that it still works without useMemo — and then add it to optimize performance.

I see that your concern is storing the instance unless normalisedPath changes. Does the normalisedPath change?

My feeling is we would be better here with class component, where we just store the instances.

lib/ts/recipe/session/recipe.ts
```diff
@@ -28,18 +28,39 @@ export default class Session extends RecipeModule<unknown, unknown, unknown, any
...
-        }
-
         sessionSdk.init({
             ...config,
-            onHandleEvent: this.onHandleEvent,
+            onHandleEvent: async (event) => {
+                if (config.onHandleEvent !== undefined) {
+                    config.onHandleEvent(event);
+                }
+                if (event.action === "SESSION_CREATED" || event.action === "REFRESH_SESSION") {
```
dulowski-marek, line 42: I'm moving this to a separate method to maintain one level of abstraction in this function.

lib/ts/recipe/session/sessionAuth.tsx
```diff
@@ -47,125 +49,96 @@ const SessionAuth: React.FC<Props> = ({ children, ...props }) => {
 
     const session = useRef(Session.getInstanceOrThrow());
 
-    const setInitialContext = useCallback(async (): Promise<void> => {
-        const sessionExists = await doesSessionExist();
+    // on mount
+    useEffect(() => {
+        let cancelUseEffect = false;
+        async function setInitialContextAndMaybeRedirect() {
```
dulowski-marek, line 55: I'm doing a slight refactor of this function to avoid so many `if..else` nesting levels and separate concerns better.

lib/ts/recipe/session/sessionAuth.tsx
```diff
@@ -47,125 +49,96 @@ const SessionAuth: React.FC<Props> = ({ children, ...props }) => {
...
-        return setContext({
-            doesSessionExist: sessionExists,
-            jwtPayload,
-            userId,
-        });
+        setInitialContextAndMaybeRedirect();
+        return () => {
+            cancelUseEffect = true;
+        };
     }, []);
```
dulowski-marek, line 98: What if the `requireAuth` prop changes? We are not going to execute this logic, despite that it changes the behaviour (user won't get redirected when this prop changes).

lib/ts/recipe/session/sessionAuth.tsx
```diff
@@ -47,125 +49,96 @@ const SessionAuth: React.FC<Props> = ({ children, ...props }) => {
...
-
-                    // ...else fallback to default behaviour
-                    if (!hasParentProvider(parentSessionContext)) {
-                        setContext({
-                            doesSessionExist: false,
-                            userId: "",
-                            jwtPayload: {},
-                        });
+                    if (props.requireAuth === true) {
+                        if (props.onSessionExpired !== undefined) {
```
dulowski-marek, line 117: It's possible to pass both `redirectToLogin` and `onSessionExpired`, while the former will be called only if `onSessionExpired` doesn't exist. This should be tested and documented.

lib/ts/recipe/session/sessionAuth.tsx
```diff
@@ -47,125 +49,96 @@ const SessionAuth: React.FC<Props> = ({ children, ...props }) => {
...
-            return;
-        }
-
-        // If the session doesn't exist and we require auth, redirect to login
-        if (context.doesSessionExist === false && props.requireAuth === true) {
-            props.redirectToLogin();
-            return;
-        }
-    }, [context, props]);
+    }, []);
```
dulowski-marek, line 136: If we don't pass dependencies here, we're going to have a stale handler (because props may change). This is fine, as this would be the behaviour of `componentDidMount`, but needs to be documented.

dulowski-marek (commented):
These are comments for my reference. I'm going to resolve them myself, so no need to act on your side.

Overall, I think that this component is way too complex, because we are trying to cover each and every scenario with it. It relies on props which are changeable by nature (like `requireAuth`), which could be communicated better by using a exposing a more granular component `<RequireAuth onUnauthorised={/* render prop */} />`.

At this moment though, to avoid API changes, I will work on resolving these.

lib/ts/recipe/session/sessionAuth.tsx
```diff
@@ -16,12 +16,14 @@
...
 import Session from "./recipe";
-import { RecipeEvent, SessionContextType } from "./types";
+import { RecipeEventWithSessionContext, SessionContextType } from "./types";
 import { doesSessionExist, getJWTPayloadSecurely, getUserId } from "./index";
 
+// if it's not the default context, it means SessionAuth from top has
+// given us a sessionContext.
 const hasParentProvider = (ctx: SessionContextType) => !isDefaultContext(ctx);
 
 type PropsWithoutAuth = {
```
dulowski-marek, line 29: Props should be adjusted as well, because when `requireAuth=true` we require either `redirectToLogin` or `onSessionExpired`

lib/ts/components/superTokensRoute.tsx
rishabhpoddar: - About point number 1: It makes sense.
- Yea.. we can get rid of the `useMemo`.

> Does the normalisedPath change?

No

> My feeling is we would be better here with class component, where we just store the instances.

How would you store it? You can't store react components in a state, and idk if it's a good idea to store then in instance variables. I think the actual problem we need to solve here is that `SuperTokensRouteWithRecipeId` is loosing state.

lib/ts/recipe/session/recipe.ts
rishabhpoddar: You really don't need to.. this refactoring serves no real purpose.. please don't over-optimize on internal code refactoring. It may introduce bugs, and it wastes time.

lib/ts/recipe/session/sessionAuth.tsx
rishabhpoddar: I think this function is fine the way it is. It calculates the context initially and acts based on what the value is. Please do not change it unless you see a bug in there.. no point in doing this type of refactor. Really no point.. Having nested `if..else` is 100% fine.

lib/ts/recipe/session/sessionAuth.tsx
rishabhpoddar: That's already done right? Why further adjustment? What benefit does it give us?

lib/ts/recipe/session/sessionAuth.tsx
rishabhpoddar: No.. this is expected behaviour.. And it's not surprising.. cause when you pass `onSessionExpired`, you expect that to be called in case of session expiry.. `redirectToLogin` is primarily meant to be used for when this component is mounted and no session exists and `requireAuth=true`.

We use `redirectToLogin` here as a fallback only. So this behaviour is intuitive and doesn't need to be documented..

lib/ts/recipe/session/sessionAuth.tsx
rishabhpoddar: I have a comment `// subscribe to events on mount` on top of the `useEffect`. We don't need to document it in the docs explicitly.. serves no purpose. You can add more comments in there if you like...

rishabhpoddar:
These are the problems that I see that actually exist:
- Unnecessary use of `useMemo`
- The component in the `Route` loosing state
- What is `onSessionExpired` and `redirectToLogin` props change? Then the [this useEffect](https://github.com/supertokens/supertokens-auth-react/pull/291/files#diff-d36242cf00898811e042afe75afaf221ee0b1dc0d6868473615fcabf03e4c275R67) will be using stale versions.
- Some unit tests not passing

lib/ts/recipe/session/sessionAuth.tsx
```diff
@@ -30,142 +32,127 @@ type PropsWithoutAuth = {
 
 type PropsWithAuth = {
     requireAuth: true;
-    redirectToLogin: () => void;
+    redirectToLogin?: () => void;
```
rishabhpoddar, line 35: No.. this is not optional.

lib/ts/recipe/session/sessionAuth.tsx
```diff
@@ -30,142 +32,127 @@ type PropsWithoutAuth = {
...
 };
 
+const noop = () => {
+    return;
+};
+
 const SessionAuth: React.FC<Props> = ({ children, ...props }) => {
-    if (props.requireAuth === true && props.redirectToLogin === undefined) {
-        throw new Error("You have to provide redirectToLogin function when requireAuth is true");
+    if (props.requireAuth === true) {
```
rishabhpoddar, line 47: This logic is wrong.

lib/ts/recipe/session/sessionAuth.tsx
```diff
@@ -30,142 +32,127 @@ type PropsWithoutAuth = {
...
     }
 
     const parentSessionContext = useContext(SessionContext);
     const [context, setContext] = useState<SessionContextType | undefined>(undefined);
 
     const session = useRef(Session.getInstanceOrThrow());
 
-    const setInitialContext = useCallback(async (): Promise<void> => {
-        const sessionExists = await doesSessionExist();
+    const redirectToLogin = props.requireAuth
```
rishabhpoddar, line 60: No need for this...

lib/ts/recipe/session/sessionAuth.tsx
```diff
@@ -30,142 +32,127 @@ type PropsWithoutAuth = {
...
     requireAuth: true;
-    redirectToLogin: () => void;
+    redirectToLogin?: () => void;
 };
 
 type Props = (PropsWithoutAuth | PropsWithAuth) & {
     onSessionExpired?: () => void;
 };
 
+const noop = () => {
```
rishabhpoddar, line 42: We don't need this.. since it `requireAuth` is `true`, the user needs to provide `redirectToLogin`

lib/ts/recipe/session/sessionAuth.tsx
rishabhpoddar: - This is meant to be run on mount, but can also run when `props` change (as long as the older event listening is unsubscribed).
- Adding `props` to the dependency array should fix it.

lib/ts/recipe/session/sessionAuth.tsx
```diff
@@ -30,142 +32,127 @@ type PropsWithoutAuth = {
...
+                doesSessionExist: true,
+                jwtPayload: await getJWTPayloadSecurely(),
+                userId: await getUserId(),
+            };
+        };
+
+        async function setInitialContextAndMaybeRedirect() {
+            // if this component is unmounting, or the context has already
+            // been set, then we don't need to proceed...
+            if (cancelUseEffect || context !== undefined) {
```
rishabhpoddar, line 95: This check is supposed to happen AFTER the `await` cause using the `await` can add a race condition..

lib/ts/recipe/session/sessionAuth.tsx
rishabhpoddar: No.. not true, cause `redirectToLogin` is meant to be used for if this component is mounted and no session exists, and `requireAuth=true`. Whilst `onSessionExpired` is for session expiry.

We use `redirectToLogin` as a fallback for session expiry in case `onSessionExpired` is not provided, which may be a little strange, but we had discussed that it was OK.

CHANGELOG.md
```diff
@@ -28,7 +28,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Breaking changes
 
--   If using auth wrapper without `requireAuth=false`, and the user signs out, then the user is automatically redirected to /auth
 -   The components override API has changed from `(DefaultComponent) => (props) => React.Element` to `({ DefaultComponent, …props }) => React.Element`.
```
rishabhpoddar, line 31: Need to add that on session expiry, the user will be redirected to the login page, unless the `onSessionExpired` callback is provided.

lib/ts/recipe/session/sessionAuth.tsx
```diff
@@ -30,142 +32,127 @@ type PropsWithoutAuth = {
...
-                        setContext({
-                            doesSessionExist: false,
-                            userId: "",
-                            jwtPayload: {},
-                        });
+                    if (props.requireAuth !== false) {
+                        setContext(event.sessionContext);
                     }
                     return;
                 case "UNAUTHORISED":
```
rishabhpoddar, line 129: This event can be fired multiple times, very fast - if there are many requests that are fired together and all yield `UNAUTHORISED`. This means that the `onSessionExpired` popup will be called many times. We want to avoid that by somehow keeping track of it has been called already or not.

I'm not sure what the best way to keep track of this is.. maybe a state, but then states update asynchronously. If this was a class component, we could have used an instance variable for this, and it would work fine.. not sure what the equivalent here would be.

lib/ts/recipe/session/sessionAuth.tsx
dulowski-marek: That's reasonable. I'll add this.

dulowski-marek:
> * Some unit tests not passing

Already fixed that.

lib/ts/recipe/session/sessionAuth.tsx
dulowski-marek: If it's designed to set some flag on parent component to display popup, then it should be idempotent, so multiple calls setting that flag to true shouldn't make a difference. I'd ignore that.

dulowski-marek: We may want to mention it in the detailed docs though.

rishabhpoddar: Yes. Makes sense when `onSessionExpired` is called. However, if it's `redirectToLogin` being called, then that will be called multiple times. So maybe we want to set some state in the SessionAuth component and then on that state change, we want to redirect.. This will make calling `redirectToLogin` idempotent

lib/ts/recipe/session/sessionAuth.tsx
dulowski-marek: It might be expected for us, but I'm thinking about the users. If I were presented with component:
`<SessionAuth requireAuth={true} onSessionExpired={handler1} redirectToLogin={handler2} />`, I wouldn't think that `onSessionExpired` will prevent calling `redirectToLogin`. That's a matter of personal opinion, again, so hard to judge. For me it's complex, just look how many discussions we had on it and how much thinking went into designing this behaviour 😅

rishabhpoddar: So `onSessionExpired` will not be called if a session doesn't exist.. then `redirectToLogin` will be called. 

As the name suggests, `onSessionExpired` will only be called when a session has expired.. which is expected.

Isn't this also intuitive for the end user?

lib/ts/recipe/session/sessionAuth.tsx
dulowski-marek: I see we have two options:
- either we instruct the user that this should be idempotent as well (set a flag to true that displays a redirection component)
- we make sure it’s called once by setting the state (which again increases the complexity of the component)

Let me know what you think

rishabhpoddar: From a user's point of view, making `redirectToLogin` idempotent is complicated.. so i prefer the second point.

rishabhpoddar: If we are doing this for `redirectToLogin`, we can apply the exact same logic for `onSessionExpired` as well.

rishabhpoddar: One thing to be careful about here is that if `requireAuth=false`, then `onSessionExpired` should be called just once whilst the session doesn't exist. But if the session then exists again, and then unauthorized again, then `onSessionExpired` should be called again.

Issue #307, opened 2021-08-25
Create an example app for using sessions without cookies.


rishabhpoddar:
- [x] Link them in the docs
- [x] post message on discord

rishabhpoddar:
I have merged the example app into the master branch of supertokens-auth-react. So the docs can now link to the relevant parts of it as necessary.

PR #313, opened 2021-08-31
Localstorage example

porcellus:
## Summary of change

Added an example that uses localstorage instead of cookies to store tokens.

## Related issues

-   #307 

## Test Plan

Calling the API in the test app without the token cookies:
image

## Documentation changes

- [] Add a section into 'Advanced customizations' that explains modification and links to the example

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`


examples/with-localstorage/api-server.js
```diff
@@ -0,0 +1,158 @@
...
+        apiDomain,
+        websiteDomain,
+    },
+    recipeList: [
+        EmailPassword.init(),
+        Session.init({
+            // CSRF protection is not necessary in this case, because there are no cookies that would be automatically sent
+            antiCsrf: "NONE",
+            override: {
+                functions: (origImpl) => {
```
rishabhpoddar, line 47: Some comments explaining how it works would be great!

rishabhpoddar (changes requested):

Issue #312, opened 2021-08-27
Provide a way to style our components in a centralized manner


porcellus:
An example use-case for this is to overwrite the default font in just one place instead of having to do it for every form/screen
Related: #309

PR #314, opened 2021-09-01
Centralized styling

porcellus:
## Summary of change

Added an extra option to the config passed to SuperTokens.init to apply styling across all components.

## Related issues

-   #312 

## Test Plan

image

## Documentation changes

supertokens/docs#106

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`


CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
+## [0.15.6]
```
rishabhpoddar, line 8: New version for 0.15.7 will have to be made as 0.15.6 was released by me after this pr was created.

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
...
+
+## Added
+
+-   Added option to apply styling to all components
```
rishabhpoddar, line 12: Also link to the github issue here

lib/ts/styles/styleContext.tsx
```diff
@@ -33,35 +34,40 @@ const StyleContext = React.createContext<NormalisedStyle>({
...
     if (styleFromInit !== undefined) {
         // Palette is a reserved word, delete it if exists.
         delete styleFromInit.palette;
 
-        Object.keys(styleFromInit).forEach((key) => [
-            (styles[key] = {
-                ...styles[key],
-                ...styleFromInit[key],
-            }),
-        ]);
```
rishabhpoddar, line 61: Why is this removed?

lib/ts/version.ts
```diff
@@ -12,5 +12,5 @@
  * License for the specific language governing permissions and limitations
  * under the License.
  */
-export const package_version = "0.15.5";
+export const package_version = "0.15.6";
```
rishabhpoddar, line 15: Need to become 0.15.7

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "supertokens-auth-react",
-    "version": "0.15.5",
+    "version": "0.15.6",
```
rishabhpoddar, line 3: 0.15.7

rishabhpoddar (changes requested):
- Consider moving the `style` object to the root config of each recipe. Cause there are recipes that don't have any UI (like the session recipe). So emailpassword would have its own global style object and so would thirdparty and thirdpartyemailpassword - this will likely result in several code changes.
- Have you manually tested that fonts can be changed via this method? And if they are, then rubik is no longer being loaded by us?

porcellus:
I added the style property to the root because then users would only have to overwrite the font in one place instead of per recipe.
Font's can be overwritten by this, but the Rubik font is still loaded. I can do that in this PR as well, I just thought it should be kept separate as they are separate issues. Should I do that in this PR?

lib/ts/styles/styleContext.tsx
porcellus: It's done by `mergedStyles = getMergedStyles(mergedStyles, styleFromInit);`

porcellus: Which upon reviewing it again a bit misleading: it actually writes into the first parameter instead of creating a new object (and it's used everywhere like it creates a new object).

rishabhpoddar:
Okay. Makes sense. You can make the font change in another PR.

lib/ts/styles/styleContext.tsx
rishabhpoddar: That's odd.. not sure why this would have been done in the first place if `getMergedStyles` existed before. hmm..

rishabhpoddar:
Also, please resolve the conflicts in this PR.

lib/ts/styles/styleContext.tsx
porcellus: Well, it was a tiny bit different because the `style` object it was merging into had an extra palette property, but I don't think that would've been an issue either. The only difference I can find is that this uses `Object.keys` and other one uses a `for..in` loop which also enumerates inherited properties.
Can I resolve this?

PR #329, opened 2021-10-27
Adding bundle size checking

porcellus:
## Summary of change

Adding bundle size checking into PRs and as a script

## Related issues

-   supertokens/supertokens-core#178

## Test Plan

Build change only

## Documentation changes

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`


porcellus:
I didn't update the version number, since the PR doesn't touch the code we release

rishabhpoddar:
- This is good. However I also prefer if there is check for this in the CICD pipeline as well - when we add a dev tag. And if it crosses a certain limit, then it should fail the CICD.
- Also, if we add a new recipe, will there need to be a change to the script to add that recipe's bundle also?

rishabhpoddar:
@porcellus Also, if we add a new recipe, will there need to be a change to the script to add that recipe's bundle also?

package.json
```diff
@@ -79,8 +81,31 @@
...
+            "path": "recipe/thirdpartyemailpassword/index.js",
+            "limit": "110kb"
+        },
+        {
+            "path": "recipe/thirdparty/index.js",
+            "limit": "110kb"
+        },
+        {
+            "path": "recipe/emailpassword/index.js",
+            "limit": "110kb"
```
rishabhpoddar, line 106: Shouldn't we have lower limits on these compared to that of thirdpartyemailpassword?

rishabhpoddar: Also, will we have to add to this everytime we add a new recipe? If yes, you should edit the PR template's checklist for this repo to say to modify this (and any other script) if we add a new recipe.

rishabhpoddar (changes requested):

PR #330, opened 2021-11-01
Added code changes to check conditions for logged-in users 

tyaga001:
Added code changes to check conditions for logged-in users should not be able to log in again in another session.


rishabhpoddar:
Hi @tyaga001 , thanks for the PR. We will each out to you via email for the next steps. For now though, I am closing this PR since (even though it works), it's not exactly the right way of doing this with SuperTokens :)

Will explain over our next call. Thanks

PR #341, opened 2021-12-07
Passwordless

porcellus:
## Summary of change

Adding implementation for the passwordless recipe

## Related issues

-   #338 

## Test Plan

Some tests are already added, but more are needed.

## Documentation changes

Done separately, entire new recipe docs

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

-   [x] Tests!
- [x] We should discuss which components are overridable and how the colour palette and CSS can be changed
-  [x] Extract common component lib from emailpassword recipe
- [x] Login checking in LinkSentScreen
- [x] Handling custom redirection when using passwordless
- [x] Disable default implementation for UI config is there?
- [x] Allow user to change input validation of phone / email?
- [x] Is deviceInfo getting saved during link code only flow as well?
- [x] Fix UI / UX  copy changes (https://github.com/supertokens/supertokens-auth-react/pull/341#issuecomment-1000116796)
- [x] Test embedding all components
- [x] Allow both phone number and email mode (depending on which user fills). 
   - [x] e2e tests for these as well.
- [x] Manually test that user signs up with email, then we add a phone number associated with them, then that user can sign in with the phone number.
- [x] Check how GENERAL_ERROR is handled for each of the API calls and forms:
   - during initial form fill
   - resending code
   - link clicked screen consumeCode
   - OTP consumeCode
- [x] How to customise UI of entering mobile number box
   - Do we need to write about this in the docs specifically?
   - Change icon for international number, or change the color of the existing icon?
- [x] If resend code / createCode API throws a general_error, we should not start the timer for resending and show the error message instead.
- [x] Edit test/with-typescript/App.tsx to have the config
- [x] Resolve / remove pending TODOs in comments
- [x] Check copy everywhere.
- [x] Resolve this: https://github.com/supertokens/supertokens-auth-react/pull/341#issuecomment-1003989409
- [x] Fix https://github.com/supertokens/supertokens-auth-react/pull/341#issuecomment-1009670513
- [x] Resolve https://github.com/supertokens/supertokens-auth-react/pull/341#issuecomment-1010824554


lib/tsconfig_dev.json
```diff
@@ -0,0 +1,24 @@
+{
```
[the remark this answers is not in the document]
rishabhpoddar: Needs to be added to .npmignore?

rishabhpoddar (changes requested):

lib/ts/recipe/passwordless/types.ts
```diff
@@ -0,0 +1,328 @@
...
+        privacyPolicyLink?: string;
+        /*
+         * Terms and conditions link for the sign-up form.
+         */
+        termsOfServiceLink?: string;
+    } & PasswordlessNormalisedBaseConfig;
+    userInputCodeForm: PasswordlessNormalisedBaseConfig;
+    linkClickedScreen: PasswordlessNormalisedBaseConfig;
+
+    contactInfoType: "PHONE" | "EMAIL";
```
rishabhpoddar, line 205: This should be called `contactMethod` - make it consistent with backend

lib/ts/recipe/passwordless/types.ts
```diff
@@ -0,0 +1,328 @@
...
+         * Terms and conditions link for the sign-up form.
+         */
+        termsOfServiceLink?: string;
+    } & PasswordlessNormalisedBaseConfig;
+
+    mobileForm: {
+        /*
+         * Must be a two-letter ISO country code (e.g.: "US")
+         */
+        defaultCountry?: CountryCode;
```
rishabhpoddar, line 192: In normalised, this should not be optional. Make it US by default?

rishabhpoddar (changes requested):

lib/ts/recipe/passwordless/recipeImplementation.ts
```diff
@@ -0,0 +1,227 @@
...
+                };
+            }
+            if ("linkCode" in input) {
+                bodyObj = {
+                    linkCode: input.linkCode,
+                    preAuthSessionId: input.preAuthSessionId,
+                };
+            }
+
+            return await querier.post("/signinup/code/consume", { body: JSON.stringify(bodyObj) }, (context) => {
```
rishabhpoddar, line 137: You need to check if this was successful. If it was, you need to fire the relevant event using `onHandleEvent`

lib/ts/recipe/passwordless/recipeImplementation.ts
```diff
@@ -0,0 +1,227 @@
...
+            input: { deviceId: string; preAuthSessionId: string } & {
+                config: NormalisedConfig;
+            }
+        ): Promise<ResendCodeApiResponse> {
+            const bodyObj = {
+                deviceId: input.deviceId,
+                preAuthSessionId: input.preAuthSessionId,
+            };
+
+            return await querier.post("/signinup/code/resend", { body: JSON.stringify(bodyObj) }, (context) => {
```
rishabhpoddar, line 99: You need to check if this was successful. If it was, you need to fire the relevant event using `onHandleEvent`

lib/ts/recipe/passwordless/recipeImplementation.ts
```diff
@@ -0,0 +1,227 @@
...
+                        status: "GENERAL_ERROR",
+                        message: validationRes,
+                    };
+                }
+                bodyObj = {
+                    phoneNumber: input.phoneNumber,
+                };
+            }
+
+            return await querier.post("/signinup/code", { body: JSON.stringify(bodyObj) }, (context) => {
```
rishabhpoddar, line 82: You need to check if this was successful. If it was, you need to fire the relevant event using `onHandleEvent`

lib/ts/recipe/passwordless/recipeImplementation.ts
```diff
@@ -0,0 +1,227 @@
...
+            | undefined
+            | {
+                  deviceId: string;
+                  preAuthSessionId: string;
+                  flowType: FlowType;
+                  contactInfo: string;
+                  contactInfoType: "EMAIL" | "PHONE";
+                  lastResend: number;
+              } {
+            const storedInfo = localStorage.getItem(PASSWORDLESS_LOGIN_ATTEMPT_INFO_STORAGE_KEY);
```
rishabhpoddar, line 188: Do not use `localstorage` directly. Instead, there is a function in ts/utils called `getLocalStorage`

lib/ts/recipe/passwordless/recipeImplementation.ts
```diff
@@ -0,0 +1,227 @@
...
+            }
+        },
+        setLoginAttemptInfo: function (input: {
+            deviceId: string;
+            preAuthSessionId: string;
+            flowType: FlowType;
+            contactInfo: string;
+            contactInfoType: "EMAIL" | "PHONE";
+        }): void {
+            localStorage.setItem(
```
rishabhpoddar, line 214: Do not use `localstorage` directly. Instead, there is a function in ts/utils called `setLocalStorage`

lib/ts/recipe/passwordless/recipeImplementation.ts
```diff
@@ -0,0 +1,227 @@
...
+                PASSWORDLESS_LOGIN_ATTEMPT_INFO_STORAGE_KEY,
+                JSON.stringify({
+                    // This can make future changes/migrations a lot cleaner
+                    version: 1,
+                    ...input,
+                })
+            );
+        },
+        clearLoginAttemptInfo: function (): void {
+            localStorage.removeItem(PASSWORDLESS_LOGIN_ATTEMPT_INFO_STORAGE_KEY);
```
rishabhpoddar, line 224: Do not use `localstorage` directly. Instead, there is a function in ts/utils called `removeFromLocalStorage`

rishabhpoddar (changes requested):

lib/ts/recipe/passwordless/utils.ts
```diff
@@ -0,0 +1,62 @@
...
+export function normalisePasswordlessConfig(config: Config): NormalisedConfig {
+    const override: any = {
+        functions: (originalImplementation: RecipeInterface) => originalImplementation,
+        components: {},
+        ...config.override,
+    };
+
+    return {
+        ...normaliseAuthRecipe(config),
+        resendCodeTimeGap: config.resendCodeTimeGap === undefined ? 900 : config.resendCodeTimeGap,
```
rishabhpoddar, line 31: COnsider renaming the config to something like `resendCodeTimeGapInSeconds` or whatever the time interval is

rishabhpoddar (changes requested):

lib/ts/recipe/passwordless/validators.ts
```diff
@@ -0,0 +1,56 @@
...
+    return undefined;
+}
+
+/*
+ * defaultValidate
+ */
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export async function defaultValidate(_: any): Promise<string | undefined> {
+    return undefined;
+}
```
rishabhpoddar, line 56: Why do we need this function?

lib/ts/recipe/passwordless/components/features/linkClickedScreen/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+            if (response.status === "GENERAL_ERROR") {
+                return this.props.recipe.redirectToAuthWithoutRedirectToPath(undefined, this.props.history, {
+                    error: "custom",
+                    message: response.message,
+                });
+            }
+
+            if (response.status === "RESTART_FLOW_ERROR") {
+                return this.props.recipe.redirectToAuthWithoutRedirectToPath(undefined, this.props.history, {
+                    error: "signin",
```
rishabhpoddar, line 69: Do you mean to send a restart event here?

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
```diff
@@ -0,0 +1,213 @@
...
+        super(props);
+
+        let error: string | undefined = undefined;
+        const errorQueryParam = getQueryParams("error");
+        if (errorQueryParam !== null) {
+            if (errorQueryParam === "signin") {
+                error = "Something went wrong. Please try again";
+            } else if (errorQueryParam === "restart") {
+                error = "The link you tried to click expired or was revoked. Please try again.";
+            }
```
rishabhpoddar, line 55: Need to handle custom error as well?

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
```diff
@@ -0,0 +1,213 @@
...
+        }
+        this.state = {
+            loaded: false,
+            loginAttemptInfo: undefined,
```
rishabhpoddar, line 59: You don't really need `loaded`. You can just check if `loginAttemptInfo !== undefined` instead of checking `if (loaded)`?

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
```diff
@@ -0,0 +1,213 @@
...
+                }
+                const res = await this.props.recipe.recipeImpl.createCode(input);
+                if (res.status === "OK") {
+                    const loginAttemptInfo = {
+                        ...res,
+                        lastResend: new Date().getTime(),
+                        contactInfoType: this.props.recipe.config.contactInfoType,
+                        contactInfo,
+                    };
+                    this.props.recipe.recipeImpl.setLoginAttemptInfo(loginAttemptInfo);
```
rishabhpoddar, line 89: Need to put await

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
```diff
@@ -0,0 +1,213 @@
...
+                    // This should never happen, but it makes TS happy
+                    throw new Error("Resend without loginAttemptInfo");
+                }
+                const res = await this.props.recipe.recipeImpl.resendCode(input);
+                if (res.status === "OK") {
+                    const loginAttemptInfo = {
+                        ...this.state.loginAttemptInfo,
+                        lastResend: new Date().getTime(),
+                    };
+                    this.props.recipe.recipeImpl.setLoginAttemptInfo(loginAttemptInfo);
```
rishabhpoddar, line 109: Should be await

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
```diff
@@ -0,0 +1,213 @@
...
+                        loginAttemptInfo,
+                    }));
+                } else if (res.status === "RESTART_FLOW_ERROR") {
+                    await this.props.recipe.recipeImpl.clearLoginAttemptInfo();
+
+                    this.setState((os) => ({
+                        ...os,
+                        error: "Please try again.",
+                        loginAttemptInfo: undefined,
+                    }));
```
rishabhpoddar, line 122: Shouldn't we redirect the user to the login page here? With an appropriate message.

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
```diff
@@ -0,0 +1,213 @@
...
+                } else if (res.status === "GENERAL_ERROR") {
+                    await this.props.recipe.recipeImpl.clearLoginAttemptInfo();
```
rishabhpoddar, line 124: Probably not a good idea to clear the login attempt info here. Maybe the general error is just some custom error by the developer

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
```diff
@@ -0,0 +1,213 @@
...
+            consumeCode: async (input) => {
+                const res = await this.props.recipe.recipeImpl.consumeCode(input);
+                if (res.status === "RESTART_FLOW_ERROR") {
+                    await this.props.recipe.recipeImpl.clearLoginAttemptInfo();
+
+                    this.setState((os) => ({
+                        ...os,
+                        error: "Please try again.",
+                        loginAttemptInfo: undefined,
+                    }));
```
rishabhpoddar, line 143: We should redirect the user to the login page here?

lib/ts/recipe/passwordless/components/features/linkClickedScreen/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+        }
+    };
+
+    render = (): JSX.Element => {
+        const componentOverrides = this.props.recipe.config.override.components;
+
+        const linkClickedScreen = this.props.recipe.config.linkClickedScreen;
+
+        const props = {
+            recipeImplementation: this.props.recipe.recipeImpl,
```
rishabhpoddar, line 92: You want to get a modified version of this that:
- on consumeCode, it clears the login attempt state if successful?
- on restart flow error, it clears the login attempt state if successful?
- anything else?

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
```diff
@@ -0,0 +1,213 @@
...
+    error: string | undefined;
+    loaded: boolean;
+    loginAttemptInfo: LoginAttemptInfo | undefined;
+};
+
+type PropType = FeatureBaseProps & {
+    recipe: Recipe;
+};
+
+class SignInUp extends PureComponent<PropType, SignInUpState> {
```
rishabhpoddar, line 44: One missing aspect is that if someone has enabled both userinput and link, and they have redeemed the link code on another tab, then this tab should change to show the "success UI" which shows that close tab cause you have already signed in text.

lib/ts/types.ts
```diff
@@ -16,8 +16,10 @@ import RecipeModule from "./recipe/recipeModule";
 import NormalisedURLPath from "./normalisedURLPath";
```
rishabhpoddar, line 16: We should discuss changes to this file on a call

rishabhpoddar (changes requested):

lib/ts/utils.ts
```diff
@@ -39,6 +39,11 @@ export function getQueryParams(param: string): string | null {
     return urlParams.get(param);
 }
 
+export function getURLHash(): string {
```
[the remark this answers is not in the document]
porcellus: It returns an empty string

lib/tsconfig_dev.json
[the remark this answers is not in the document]
porcellus: It's not needed, it was just useful during development and I committed it by accident. We can keep it if it makes things easier for anyone else, then I'll add it to `.npmignore`

package.json
```diff
@@ -108,6 +111,10 @@
         {
             "path": "recipe/emailpassword/index.js",
             "limit": "100kb"
+        },
+        {
+            "path": "recipe/passwordless/index.js",
+            "limit": "100kb"
```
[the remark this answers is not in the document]
porcellus: I wanted to update this once the PR is mostly done, but I'll just set it to 160 for now.

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
porcellus: It's used to make sure we don't pop the email/mobile form in&out before displaying the OTP form when refreshing the page.

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
porcellus: This kind of error is too easy to make, I think we should a [this rule](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-floating-promises.md) to the linter. I'll do this later because it reports this in a few other places (where they are hopefully intentional)

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
porcellus: By removing the loginAttemptInfo you get back to the original mobile or email form.

lib/ts/recipe/passwordless/components/features/linkClickedScreen/index.tsx
porcellus: I don't think the restart flow error should clear because it would block you from entering the user input code in the combined flow, which may be a use-case if the user noticed copying the link didn't work and tries to just type it in.

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
porcellus: That's one of the TODOs, but I assumed we would just redirect the same as a successful login would. Should I add a separate screen for this? Maybe we could add an extra field to the redirection context?

lib/tsconfig_dev.json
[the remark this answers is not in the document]
rishabhpoddar: We can keep it. But what made it useful during dev?

lib/ts/recipe/passwordless/types.ts
rishabhpoddar: @porcellus missed this comment?

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
rishabhpoddar: Let's discuss this:
- It's not obvious that this is happening, cause the recipeImplementation.ts file isn't doing that
- What will be the error message displayed to the user in that case?

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
rishabhpoddar: Let's discuss this over a call.

rishabhpoddar (changes requested):

lib/tsconfig_dev.json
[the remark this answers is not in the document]
rishabhpoddar: You can update the npm run watch command to use this tsconfig.

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
```diff
@@ -0,0 +1,214 @@
...
+                        ...os,
+                        error: undefined,
+                        loginAttemptInfo,
+                    }));
+                } else if (res.status === "RESTART_FLOW_ERROR") {
+                    await this.props.recipe.recipeImpl.clearLoginAttemptInfo();
+
+                    this.setState((os) => ({
+                        ...os,
+                        error: "Please try again.",
```
rishabhpoddar, line 123: We need to double check the copy everywhere.

rishabhpoddar (changes requested):

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
rishabhpoddar: We have a separate screen for this. So please implement that.

lib/ts/recipe/passwordless/recipe.tsx
```diff
@@ -81,15 +83,33 @@ export default class Passwordless extends AuthRecipe<
...
 
         return features;
     };
 
     getDefaultRedirectionURL = async (context: GetRedirectionURLContext): Promise<string> => {
+        if (context.action === "SUCCESS_IN_ANOTHER_TAB") {
+            const resetPasswordPath = new NormalisedURLPath(DEFAULT_SUCCESS_PATH);
+            return `${this.config.appInfo.websiteBasePath.appendPath(resetPasswordPath).getAsStringDangerous()}?rid=${
+                this.config.recipeId
+            }`;
```
rishabhpoddar, line 104: Why reset password?

lib/ts/recipe/passwordless/recipe.tsx
```diff
@@ -81,15 +83,33 @@ export default class Passwordless extends AuthRecipe<
                 component: (props) => this.getFeatureComponent("linkClickedScreen", props),
             };
         }
+        if (this.config.linkClickedScreen.disableDefaultImplementation !== true) {
+            const normalisedFullPath = this.config.appInfo.websiteBasePath.appendPath(
+                new NormalisedURLPath("/success")
```
rishabhpoddar, line 88: If we must make this its own page, I don't think `/success` is an appropriate path. Perhaps something like `/close-tab`.

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
```diff
@@ -161,7 +164,29 @@ class SignInUp extends PureComponent<PropType, SignInUpState> {
...
+        const checkSessionIntervalHandle = setInterval(async () => {
+            // TODO: check if we can call into the session recipe from here.
+            const hasSession = await Session.doesSessionExist();
+            if (hasSession) {
+                return this.props.recipe.redirect(
+                    {
+                        action: "SUCCESS_IN_ANOTHER_TAB",
+                    },
+                    this.props.history
+                );
```
rishabhpoddar, line 179: I don't think it makes sense to make this its own page. I say this cause it implies then that someone can navigate to it and see the UI there, which doesn't make sense from a user flow point of view. I'd rather have it as another component within this feature component (just how enter phone, and enter user input code are).

That being said, i'd still keep the user config and overridability of this component as is in this PR.

rishabhpoddar (changes requested):

test/server/index.js
```diff
@@ -39,6 +58,19 @@ const WEB_PORT = process.env.WEB_PORT || 3031;
...
+function saveCode({ email, phoneNumber, preAuthSessionId, urlWithLinkCode, userInputCode }) {
+    const device = deviceStore.get(preAuthSessionId) || {
+        preAuthSessionId,
+        codes: [],
+    };
+    device.codes.push({
+        urlWithLinkCode,
+        userInputCode,
+    });
+    console.log({ email, phoneNumber, urlWithLinkCode, userInputCode });
```
rishabhpoddar, line 71: You should remove this console.log and others from this file if they are not needed.

rishabhpoddar (changes requested):

test/server/index.js
```diff
@@ -143,11 +109,21 @@ app.get("/ping", async (req, res) => {
...
+    if (req.body && req.body.configUpdates) {
+        for (const update of req.body.configUpdates) {
+            await setKeyValueInConfig(update.key, update.value);
+        }
+    }
     let pid = await startST();
     res.send(pid + "");
 });
 
 app.post("/beforeeach", async (req, res) => {
```
rishabhpoddar, line 121: When testing with a backend SDK, there are two servers:
- `TEST_SERVER_BASE_URL`: This is a node server in the auth-react package. This is in charge of starting / stopping ST etc...
- `TEST_APPLICATION_SERVER_BASE_URL`: This is the node server in the supertokens-node package. Or a go server in supertokens-golang package. This is where the for-tests app will query. Therefore, these operations below will need to be carried out in this server while testing. This in turn implies that the frontend will have to query the `/beforeeach` route of the `TEST_APPLICATION_SERVER_BASE_URL` as well - so this will require a change in the `before(async function () {` block of the frontend test files.

test/server/index.js
```diff
@@ -143,11 +109,21 @@ app.get("/ping", async (req, res) => {
...
+    users = [];
+    flowTypes = new Map();
+    deviceStore = new Map();
+    maximumCodeInputAttempts = 3;
```
rishabhpoddar, line 125: Out of these vars, only deviceStore is actually defined.

test/server/index.js
```diff
@@ -190,10 +166,39 @@ app.get("/token", async (_, res) => {
...
+    res.send({
+        available,
+    });
+});
+
 app.use(errorHandler());
 
 app.use(async (err, req, res, next) => {
     try {
+        console.error(err);
```
rishabhpoddar, line 201: Remove console.log?

rishabhpoddar (changes requested):

test/server/index.js
```diff
@@ -190,10 +166,39 @@ app.get("/token", async (_, res) => {
...
         res.sendStatus(500).send(err);
```
rishabhpoddar, line 202: Should be `res.status(500).send(err)`

rishabhpoddar (changes requested):

CHANGELOG.md
```diff
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [0.18.0] - 2021-12-08
```
rishabhpoddar, line 8: Not sure why we need to release this as a breaking change. Furthermore, please revert the changes of the example/with-* apps to not use 0.18.0 for auth-react since that is something we will update after release of passwordless.

rishabhpoddar (changes requested):

lib/ts/recipe/passwordless/types.ts
```diff
@@ -0,0 +1,341 @@
...
+        components: ComponentOverrideMap;
+    };
+} & NormalisedAuthRecipeModuleConfig<GetRedirectionURLContext, PreAPIHookContext, OnHandleEventContext>;
+
+export type Config = UserInput &
+    AuthRecipeModuleConfig<GetRedirectionURLContext, PreAPIHookContext, OnHandleEventContext>;
+
+export type PasswordlessFeatureBaseConfig = {
+    disableDefaultImplementation?: boolean;
+} & FeatureBaseConfig;
```
rishabhpoddar, line 224: This is not being used anywhere?

rishabhpoddar (changes requested):

CHANGELOG.md
rishabhpoddar: @porcellus did you miss this comment?

porcellus: I did sorry. I'll revert them and add it as a patch version. Although according to semver we are supposed to be increasing the minor version, I guess we are not breaking anything and we are in the 0.X territory anyway.

rishabhpoddar: Actually.. This is a new "big" feature. So even if it's not breaking, perhaps it makes sense to keep it this way..?

lib/ts/recipe/session/recipe.ts
```diff
@@ -40,7 +40,7 @@ export default class Session extends RecipeModule<unknown, unknown, unknown, any
                     config.onHandleEvent(event);
                 }
 
-                this.notifyListeners(event);
+                void this.notifyListeners(event);
```
rishabhpoddar: Why are we adding `void` in here and other places?

lib/ts/styles/styleContext.tsx
```diff
@@ -74,7 +74,8 @@ export function StyleProvider({
  * Helpers
  */
 function getMergedPalette(defaultPalette: NormalisedPalette, rawPalette: Record<string, string>): NormalisedPalette {
-    const palette = defaultPalette;
+    // We copy the defaultPalette into in order to not update the original
+    const palette = { colors: { ...defaultPalette.colors }, fonts: { ...defaultPalette.fonts } };
```
rishabhpoddar: What if the `defaultPalette` has something new later? Why not do a generic deep clone?

lib/ts/recipe/passwordless/recipeImplementation.ts
```diff
@@ -0,0 +1,275 @@
...
+                    deviceId: info.deviceId,
+                    flowType: info.flowType,
+                    preAuthSessionId: info.preAuthSessionId,
+                    lastResend: info.lastResend,
+                };
+            } catch (ex) {
+                return undefined;
+            }
+        },
+        setLoginAttemptInfo: function (input: {
```
rishabhpoddar: You also need to store redirectToPath (optionally) in here? So that custom redirection via magic link flow works too.

lib/ts/recipe/passwordless/components/features/linkClickedScreen/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+            if (response.status === "RESTART_FLOW_ERROR") {
+                return this.props.recipe.redirectToAuthWithoutRedirectToPath(undefined, this.props.history, {
+                    error: "restart_link",
+                });
+            }
+            if (response.status === "OK") {
+                await this.props.recipe.recipeImpl.clearLoginAttemptInfo();
+                return this.props.recipe.redirect(
+                    { action: "SUCCESS", isNewUser: response.createdUser, redirectToPath: getRedirectToPathFromURL() },
```
rishabhpoddar: - The magic link won't have the custom redirect path. You need to save this in localstorage along with device info and fetch it from there.
- You should add a test for testing custom redirection during link flow and user input code flow

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
```diff
@@ -0,0 +1,257 @@
...
+                }
+                const res = await this.props.recipe.recipeImpl.createCode(input);
+                if (res.status === "OK") {
+                    const loginAttemptInfo = {
+                        ...res,
+                        lastResend: new Date().getTime(),
+                        contactMethod: this.props.recipe.config.contactMethod,
+                        contactInfo,
+                    };
+                    await this.props.recipe.recipeImpl.setLoginAttemptInfo(loginAttemptInfo);
```
rishabhpoddar: Why not call `setLoginAttemptInfo` inside the `recipeImpl.createCode` itself?

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
```diff
@@ -0,0 +1,257 @@
...
+                    // This should never happen, but it makes TS happy
+                    throw new Error("Resend without loginAttemptInfo");
+                }
+                const res = await this.props.recipe.recipeImpl.resendCode(input);
+                if (res.status === "OK") {
+                    const loginAttemptInfo = {
+                        ...this.state.loginAttemptInfo,
+                        lastResend: new Date().getTime(),
+                    };
+                    await this.props.recipe.recipeImpl.setLoginAttemptInfo(loginAttemptInfo);
```
rishabhpoddar: Why not call `setLoginAttemptInfo` inside the `recipeImpl.resendCode` itself?

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
```diff
@@ -0,0 +1,257 @@
...
+                    this.setState((os) => ({
+                        ...os,
+                        error: undefined,
+                        loginAttemptInfo,
+                    }));
+                } else if (res.status === "RESTART_FLOW_ERROR") {
+                    await this.props.recipe.recipeImpl.clearLoginAttemptInfo();
```
rishabhpoddar: Why not call `clearLoginAttemptInfo` inside `recipeImpl.resendCode` itself?

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
```diff
@@ -0,0 +1,257 @@
...
+                        loginAttemptInfo: undefined,
+                    }));
+                } else if (res.status === "GENERAL_ERROR") {
+                    this.setState((os) => ({
+                        ...os,
+                        error: res.message,
+                        loginAttemptInfo: undefined,
+                    }));
+                } else if (res.status === "OK") {
+                    await this.props.recipe.recipeImpl.clearLoginAttemptInfo();
```
rishabhpoddar: Why not call `clearLoginAttemptInfo` inside `recipeImpl. consumeCode` itself?

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
```diff
@@ -0,0 +1,257 @@
...
+                        loginAttemptInfo: undefined,
+                    }));
+                }
+                return res;
+            },
+
+            consumeCode: async (input) => {
+                const res = await this.props.recipe.recipeImpl.consumeCode(input);
+                if (res.status === "RESTART_FLOW_ERROR") {
+                    await this.props.recipe.recipeImpl.clearLoginAttemptInfo();
```
rishabhpoddar: Why not call `clearLoginAttemptInfo` inside `recipeImpl. consumeCode` itself?

lib/ts/recipe/passwordless/components/features/linkClickedScreen/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+                });
+            }
+
+            if (response.status === "RESTART_FLOW_ERROR") {
+                return this.props.recipe.redirectToAuthWithoutRedirectToPath(undefined, this.props.history, {
+                    error: "restart_link",
+                });
+            }
+            if (response.status === "OK") {
+                await this.props.recipe.recipeImpl.clearLoginAttemptInfo();
```
rishabhpoddar: Why not call `clearLoginAttemptInfo` inside `recipeImpl. consumeCode` itself?

lib/ts/recipe/passwordless/components/features/linkClickedScreen/index.tsx
rishabhpoddar: However, if the user mistyped the user input code incorrectly a bunch of times, that would result in the flow to restart (which would prevent the magic link from working even if used correctly - is that correct?)

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
```diff
@@ -0,0 +1,257 @@
...
+        };
+    };
+
+    componentDidMount = async (): Promise<void> => {
+        const loginAttemptInfo = await this.getModifiedRecipeImplementation().getLoginAttemptInfo();
+
+        // We could be using storage events for this, but we need to keep customization in mind.
+        // Someone could be using something else other than localstorage.
+        const checkSessionIntervalHandle = setInterval(async () => {
+            // TODO: check if we can call into the session recipe from here.
```
rishabhpoddar: You can remove this TODO. It's safe to assume that we can call the session recipe.

lib/ts/recipe/passwordless/index.ts
```diff
@@ -0,0 +1,68 @@
...
+        } else {
+            if (input.redirectBack === false || input.redirectBack === undefined) {
+                return Passwordless.getInstanceOrThrow().redirectToAuthWithoutRedirectToPath(input.show);
+            } else {
+                return Passwordless.getInstanceOrThrow().redirectToAuthWithRedirectToPath(input.show);
+            }
+        }
+    }
+
+    static PasswordlessAuth = PasswordlessAuth;
```
rishabhpoddar: We also need to export the SignInUp feature component as is done in other recipes. Like this:
```js
static SignInAndUp = (prop?: any) => Passwordless.getInstanceOrThrow().getFeatureComponent("signInUp", prop);
```

rishabhpoddar (changes requested):

lib/ts/recipe/passwordless/utils.ts
```diff
@@ -0,0 +1,64 @@
...
+        },
+
+        mobileForm: {
+            ...config.mobileForm,
+            ...normalisePasswordlessBaseConfig(config.mobileForm),
+        },
+        userInputCodeForm: normalisePasswordlessBaseConfig(config.userInputCodeForm),
+        linkClickedScreen: normalisePasswordlessBaseConfig(config.linkClickedScreen),
+
+        contactMethod: config.contactMethod,
```
rishabhpoddar: You need to check if the user did not provide `contactMethod`, it should throw an error. Cause in JS implementation, I can do `Passwordless.init({})`, and it compiles and shows the phone number based impl by default.

rishabhpoddar (changes requested):

lib/ts/recipe/passwordless/recipeImplementation.ts
```diff
@@ -0,0 +1,275 @@
...
+            input: ({ email: string } | { phoneNumber: string } | { deviceId: string; preAuthSessionId: string }) & {
+                config: NormalisedConfig;
+            }
+        ): Promise<CreateCodeApiResponse> {
+            let bodyObj;
+            if ("email" in input) {
+                const validationRes = await input.config.validateEmailAddress(input.email);
+                if (validationRes !== undefined) {
+                    return {
+                        status: "GENERAL_ERROR",
```
rishabhpoddar: If we return a `GENERAL_ERROR`, it shows errors in two places - one below the input field, and one on top

rishabhpoddar (changes requested):

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
porcellus: We do not want to clear on `RESTART_FLOW_ERROR`: in the OTP + Link flow we can still continue if consuming the linkCode fails. We can check which the flow and what kind of consume ran and do the cleanup based on that, but I thought this kind of logic was supposed to be in the feature component, so the user could override it.

lib/ts/recipe/passwordless/components/features/linkClickedScreen/index.tsx
porcellus: That's correct. In this case the user input code part of the flow should do the cleanup. Thinking about this in more general terms, we shouldn't really rely on expired/invalid devices staying in DB, because the last allowed attempt will delete the device as well, so other attempts to consume either the link or the user input code would only see that the device doesn't exists.
If this becomes a problem in the future we could have a separate endpoint where the UI could check if the login attempt is still valid (polling by deviceIdHash)

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
porcellus: This has been changed since then, can I resolve this comment? (Maybe we could add a checkmark into the PR todos section for the UX/copy check if necessary)

CHANGELOG.md
porcellus: I think it makes more sense that way. Should I move it to 0.19.0?

examples/for-tests/src/App.js
```diff
@@ -501,6 +514,95 @@ function getEmailPasswordConfigs({ disableDefaultImplementation }) {
...
+        },
+        mobileForm: {
+            disableDefaultImplementation,
+            style: theme.style,
+
+            privacyPolicyLink: "https://supertokens.io/legal/privacy-policy",
+            termsOfServiceLink: "https://supertokens.io/legal/terms-and-conditions",
+        },
+        userInputCodeForm: {
+            disableDefaultImplementation,
```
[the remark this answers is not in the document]
porcellus: It is, it's not a separate screen/feature component. I'll add the comment.

lib/ts/recipe/passwordless/types.ts
```diff
@@ -0,0 +1,341 @@
...
+
+export type UserInput = {
+    contactMethod: "PHONE" | "EMAIL";
+
+    validateEmailAddress?: (email: string) => Promise<string | undefined> | string | undefined;
+    validatePhoneNumber?: (phoneNumber: string) => Promise<string | undefined> | string | undefined;
+
+    resendCodeTimeGapInSeconds?: number;
```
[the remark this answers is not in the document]
porcellus: It's also applicable in the `linkSent`.

lib/ts/recipe/session/recipe.ts
porcellus: `void` signifies that we are intentionally ignoring the return value of `notifyListeners`, which in this case is a Promise that we do not want to await/handle. This way we can still have these "floating promises" and not trigger the lint rule.

lib/ts/styles/styleContext.tsx
porcellus: We'd have to add it here, but (and update the same kind of thing in a bunch of other places). We didn't have a generic deep copy util, so I didn't want to add one just for this case, because it changes very rarely.

rishabhpoddar:
# UX / copy / design feedback (in progress):
- [ ] In email password, the place holder for Email input is “Email address”. But here, it’s youremail@example.com.
- [ ] "Resend in ___" text is clickable. It should not be
- [ ] If the entered email is really long (edge case), then the UI for enter OTP shows like this (same goes for magic link sent screen):
Screenshot 2021-12-23 at 13 12 30
- [ ] Change "This login attempt expired or was revoked. Please try again." to "Login timed out. Please try again"
- [ ] Change "The link you tried to click expired or was revoked. Please try again."  to "The link you clicked expired or was revoked. Please try again."
- [ ] Displaying attempts left has a leading "0". Like "Attempts left: 04". Should be just "Attempts left: 4".
- [ ] If the user has entered the wrong OTP, we must clear that OTP field.
- [ ] This is a design issue, but if the user has both link and OTP flow, then the text in the enter OTP screen should say "An OTP and magic link was sent to you at <EMAIL>" (likewise for mobile based login)
- [ ]

porcellus:
>  "Resend in ___" text is clickable. It should not be

I'll change the cursor.

> If the entered email is really long (edge case), then the UI for enter OTP shows like this (same goes for magic link sent screen):

I think we should add ellipses here or would you prefer breaking the email into multiple lines?

>  We should not display the attempts left in case of wrong OTP entered.

I'm happy to change this but it was in some of the designs.

> That is incorrect. It should instead say something like "Login timed out. Please try again"?

I think the message should include both, since we can't reliably tell the difference.

> If the frontend thinks that the phone number is valid, but the backend doesn't think that, ...

Hmm, maybe we could just have the two line up more or return a FieldError in this case instead of a GeneralError.

> I don't think that the tick is needed at all

I can remove it, but it was in the design and the email input has it, so it's a bit more consistent this way. Also, while not necessary it's good feedback that the input recognized your phone number and you didn't mistype/forget a digit.

> If I enter my email and then see the enter OTP screen. Then I close the tab and come back sometime later. Then I will still see the enter OTP screen. Which can be confusing if I had forgotten that I had entered my email. Perhaps we should use something like session storage instead of localstorage to store the device info data?

I think we could use sessionstorage, that'd work fine, since it still allows refreshing to load the same page but opening (or re-opening) another tab would get you back to the normal login screen. Also, this is related to the next case.

>  If I enter an email, then see the enter OTP screen. I then shut the tab, and come back to it after sometime (after the resend timer has finished), then I do not get a new email on its own. That's weird cause the screen says "An OTP was sent to you.."

Hmm, but the OTP was sent, just a bit earlier... I mean reopening the page like that can be a confusing, but resending isn't really a solution I think. I'd not expect an action like that on pageload in an SPA. 
We could add the time it was sent to the message, but I'm not sure if that's a solution.

rishabhpoddar:
> I think we should add ellipses here or would you prefer breaking the email into multiple lines?

Elipses is good

> I'm happy to change this but it was in some of the designs.

Yes. Please don't show the attempts left. The reason is cause if someone doesn't want this feature of limiting OTP attempts, they can change the number in the core to a very high value. And if we show this, then the high number will look weird.

> I think the message should include both, since we can't reliably tell the difference.

Okay. So what will the new message be?

> Hmm, maybe we could just have the two line up more or return a FieldError in this case instead of a GeneralError.

No need. I don't want to make the API more complex cause of this.

> I can remove it, but it was in the design and the email input has it, so it's a bit more consistent this way. Also, while not necessary it's good feedback that the input recognized your phone number and you didn't mistype/forget a digit.

So the tick doesn't show if I type the right number / email cause when I click next, it takes me to the next screen very quickly anyway. It only shows if I loose focus on the input text (Which is unlikely cause there is only one input text), or if the frontend thinks the number is correct, but the backend disagrees - which is then weird to show the tick.

>  but opening (or re-opening) another tab would get you back to the normal login screen.

Only if the browser was closed completely before opening another tab? We need to understand clearly the semantics of session storage vs localstorage.

> Hmm, but the OTP was sent, just a bit earlier... I mean reopening the page like that can be a confusing, but resending isn't really a solution I think. I'd not expect an action like that on pageload in an SPA.
We could add the time it was sent to the message, but I'm not sure if that's a solution.

Cool. No need to make any changes related to this then.

porcellus:
> If the passwordless_max_code_input_attempts is 1...

We don't have access to this in general on the UI, only if the user can try again. We could add extra logic to see if this was the first attempt, but even the first attempt could be an expired OTP, which would produce the same RESTART_FLOW_ERROR.

> Okay. So what will the new message be?

I'll just add "Login timed out or ...". Btw, should these messages say login? Login or signup would look weird, but that's technically the correct message.

>  If we enter an invalid phone on the frontend, we see "Phone number is invalid" below the input text and on the top of the form. We should only see below the input text

I think this happens because the validation errors are not cleared when calling the API and an earlier error was stuck there. At least that's the only way I could reproduce it. I've fixed this, but I'm not sure if you got it the same way.

> session storage vs localstorage.

https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage
Some important points:
- It survives restores (closing the tab or browser then restoring it e.g.: reopening the tabs when chrome starts up again)
- Duplicate tab copies the state, so it's not strictly on one tab only.
- Opening the app in a new tab will have different session storage state
I'm not sure if switching to session storage would be a full solution. It solves the problem of closing and later opening the app (but not re-opening the tab). We could implement the polling endpoint which could redirect to the auth page if the `preAuthSessionId` expired, but I'm not sure if that's necessary in v1.

rishabhpoddar:
> We don't have access to this in general on the UI, only if the user can try again. We could add extra logic to see if this was the first attempt, but even the first attempt could be an expired OTP, which would produce the same RESTART_FLOW_ERROR.

Ahh i see.. Then the message should be something generic then. Like "Login unsuccessful. Please try again". 

> I'll just add "Login timed out or ...". Btw, should these messages say login? Login or signup would look weird, but that's technically the correct message.

Saying login is fine.

> I think this happens because the validation errors are not cleared when calling the API and an earlier error was stuck there. At least that's the only way I could reproduce it. I've fixed this, but I'm not sure if you got it the same way.

Thinking about this error showing, we can eliminate showing any errors below the input text box cause it's just one input text. So all errors can be shown above the form and can be returned via "GENERAL_ERROR"

--------
About using session storage, since it's not persisted across tabs, we should continue to use localstorage (since magic links will always open in another tab). So the issue I described can remain for now I guess. We can add polling later to make this flow better - we should add a new issue for enhancing passwordless with all our ideas. Please do so.

rishabhpoddar:
## More UX feedback:
- [ ] When returning a GENERAL_ERROR output with a custom message from the enter OTP screen (during consumeCode), the user is shown the message, but the OTP screen goes away and they are shown the enter email screen again (which should not happen) - and then when I refresh the page, I see the enter OTP screen.
- [ ] When returning a GENERAL_ERROR output with a custom message from the enter OTP screen (during resendCode), the user is shown the message, but the OTP screen goes away and they are shown the enter email screen again (which should not happen) - and then when I refresh the page, I see the enter OTP screen.
- [ ] Default Resend interval is too long. It should be 1 min by default
- [ ] When both link and user input code are enabled, sending a GENERAL_ERROR with a custom message from the /verify screen, the user is shown the custom message, but they are shown the enter OTP screen - this is a bit weird - but I'm not sure what a good alternative it. So perhaps this is good enough?
- [ ] When only link is enabled, sending a GENERAL_ERROR with a custom message from the /verify screen, the user is shown the custom message, but they are shown the link sent screen - this is a bit weird - but I'm not sure what a good alternative it. So perhaps this is good enough?
- [ ] When clicking "change email" or "change phone number" button, and there is an error query param, that error query param should go away when we show the enter email / phone number screen. In general, the error query param approach is suitable for when we want to send an error message to display across screens / features. The use of this makes sense in the context of link clicked feature wanting to shown an error in the sign in up feature, but, making use of error query param within a feature component is strange.

rishabhpoddar:
## TODO (to discuss with everyone else in the team)
- [ ]  If the user has entered the wrong OTP, we must clear that OTP field - should we do this or not?

porcellus:
[ ] We've decided to split the input typings based on `contactMethod` to reduce the chance of confusing the different form configs (if we are only allowing phone numbers then the config for the form that accepts both email and phone wouldn't be applied)

porcellus:
## Proposed config changes after adding `email_or_phone contact` method:

```ts
export type SignInUpFeatureConfigInput = {
    disableDefaultImplementation?: boolean;
    resendCodeTimeGapInSeconds?: number;

    /*
     * Privacy policy link for the sign-up form.
     */
    privacyPolicyLink?: string;
    /*
     * Terms and conditions link for the sign-up form.
     */
    termsOfServiceLink?: string;

    emailOrPhoneFormStyle?: Styles;
    userInputCodeFormStyle?: Styles;
    linkSentScreenStyle?: Styles;
    linkClickedScreenStyle?: Styles;
}

export type UserInput = (
    | {
          contactMethod: "EMAIL";

          validateEmailAddress?: (email: string) => Promise<string | undefined> | string | undefined;

          signInUp: SignInUpFeatureConfigInput;
      }
    | {
          contactMethod: "PHONE";

          validatePhoneNumber?: (phoneNumber: string) => Promise<string | undefined> | string | undefined;

          signInUp: SignInUpFeatureConfigInput & {
              /*
               * Must be a two-letter ISO country code (e.g.: "US")
               */
              defaultCountry?: CountryCode;
          };
      }
    | {
          contactMethod: "EMAIL_OR_PHONE";

          validateEmailAddress?: (email: string) => Promise<string | undefined> | string | undefined;
          validatePhoneNumber?: (phoneNumber: string) => Promise<string | undefined> | string | undefined;

          signInUp: SignInUpFeatureConfigInput;
      }
) & {
    override?: {
        functions?: (originalImplementation: RecipeInterface) => RecipeInterface;
        components?: ComponentOverrideMap;
    };
} & AuthRecipeModuleUserInput<GetRedirectionURLContext, PreAPIHookContext, OnHandleEventContext>;
```

rishabhpoddar:
Flow for email and phone input box:
- When the user clicks on continue:
   - if it's an email -> continue with email.
   - else if it's a value phone number (with international code) -> continue with that.
   - else if it's a valid phone number without the international code -> change UI to show the country code dropdown -> when the field is cleared, then remove the country code drop down (back to step 1 here).

-----------
- [ ] Is it possible to tell the country code based on the number?

lib/ts/recipe/passwordless/components/themes/signInUp/phoneForm.tsx
```diff
@@ -35,11 +36,11 @@ export const MobileForm = withOverride(
                         id: "phoneNumber",
                         label: "Your Phone Number",
                         inputComponent: phoneNumberInputWithInjectedProps({
-                            defaultCountry: props.config.mobileForm.defaultCountry,
+                            defaultCountry: props.config.signInUpFeature.defaultCountry,
                         }),
                         optional: false,
                         placeholder: "+1 (389) 848-293-234",
```
rishabhpoddar, line 42: Need to change this place holder.

lib/ts/recipe/passwordless/index.ts
```diff
@@ -49,6 +50,12 @@ export default class Wrapper {
     }
 
     static PasswordlessAuth = PasswordlessAuth;
+
+    static SignInAndUp = (prop?: any) => Passwordless.getInstanceOrThrow().getFeatureComponent("signInUp", prop);
+    static SignInAndUpTheme = SignInUpThemeWrapper;
+
+    static LinkClicked = (prop?: any) =>
+        Passwordless.getInstanceOrThrow().getFeatureComponent("linkClickedScreen", prop);
 }
```
rishabhpoddar, line 59: We need to export all these as part of the non-default export too

lib/ts/recipe/passwordless/utils.ts
```diff
@@ -13,52 +13,138 @@
...
+import {
+    defaultPhoneNumberValidator,
+    defaultPhoneNumberValidatorForCombinedInput,
+    defaultEmailValidator,
+    defaultEmailValidatorForCombinedInput,
+} from "./validators";
 
 export function normalisePasswordlessConfig(config: Config): NormalisedConfig {
+    if (!["EMAIL", "PHONE", "EMAIL_OR_PHONE"].includes(config.contactMethod)) {
+        throw new Error("Please pass one of 'PHONE' or 'EMAIL' as the contactMethod");
```
rishabhpoddar, line 31: Change message to `"Please pass one of 'PHONE', 'EMAIL' or 'EMAIL_OR_PHONE' as the contactMethod"`

rishabhpoddar (changes requested):

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
```diff
@@ -160,12 +157,6 @@ class SignInUp extends PureComponent<PropType, SignInUpState> {
                         error: SIGN_IN_UP_CODE_CONSUME_RESTART_FLOW_ERROR,
                         loginAttemptInfo: undefined,
                     }));
-                } else if (res.status === "GENERAL_ERROR") {
-                    this.setState((os) => ({
-                        ...os,
-                        error: res.message,
-                        loginAttemptInfo: undefined,
-                    }));
```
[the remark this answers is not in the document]
porcellus: It's handled by `FormBase`

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
```diff
@@ -94,11 +91,13 @@ class SignInUp extends PureComponent<PropType, SignInUpState> {
                 }
                 const res = await this.props.recipe.recipeImpl.createCode(input);
                 if (res.status === "OK") {
+                    const contactMethod: "EMAIL" | "PHONE" = "email" in input ? "EMAIL" : "PHONE";
```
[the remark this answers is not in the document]
porcellus: This is about the contact method we used - I was debating renaming this to `usedContactMethod`. Would that make it clearer?

lib/ts/recipe/passwordless/components/features/linkClickedScreen/index.tsx
```diff
@@ -70,9 +70,15 @@ class LinkClickedScreen extends PureComponent<PropType, unknown> {
                 });
             }
             if (response.status === "OK") {
+                const loginAttemptInfo = await this.props.recipe.recipeImpl.getLoginAttemptInfo();
                 await this.props.recipe.recipeImpl.clearLoginAttemptInfo();
+                const pathFromUrl = getRedirectToPathFromURL();
```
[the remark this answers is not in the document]
porcellus: It might be defined - devs might want to set this information on the backend while generating the link. I'm unsure if this is a use-case, but I wanted to keep the option open.

lib/ts/recipe/passwordless/components/themes/signInUp/emailOrPhoneForm.tsx
```diff
@@ -0,0 +1,112 @@
...
+                        id: "emailOrPhone",
+                        label: "Your Email address or Phone number",
+                        inputComponent: phoneNumberVal
+                            ? phoneNumberInputWithInjectedProps({
+                                  defaultCountry: props.config.signInUpFeature.defaultCountry || "HU",
+                                  initialValue: phoneNumberVal,
+                              })
+                            : undefined,
+                        optional: false,
+                        placeholder: "+1 (389) 848-293-234",
```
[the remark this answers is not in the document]
porcellus: I updated this to the "standard", but having something that tells the user they need to enter the international number may be useful

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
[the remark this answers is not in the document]
rishabhpoddar: I'm not sure what you mean by "contact method we used". But if `"EMAIL" | "PHONE"` is intentional, then you should write a comment there explaining that this is intentional and why we do not consider `PHONE_OR_EMAIL` in here.

lib/ts/recipe/passwordless/components/features/linkClickedScreen/index.tsx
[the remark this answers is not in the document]
rishabhpoddar: For now, this is not a use case. We can add this as a use case to third party and to passwordless in another PR. For now, let's keep it consistent.

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
rishabhpoddar: Added a checklist item to check copy

CHANGELOG.md
rishabhpoddar: It should be moved to 0.18.0 since the latest released is 0.17.X

lib/ts/styles/styleContext.tsx
rishabhpoddar: Okay

rishabhpoddar:
# New UX feedback:
- [ ] Change "Your Email address or Phone number:" to "Email or Phone number:"
- [ ] Change place holder to "Email or Phone number"
- [ ] When entering email and clicking on continue, we see "Resend" button first which then quickly changes to "Resend in..". This UI glitch is strange.
- [ ] Change "The link you clicked expired or was revoked. Please try again." to "Invalid magic link. Please try again".

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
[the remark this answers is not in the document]
porcellus: Contact method we used to deliver the login info. Even if both are allowed, we only used one. I'll add this as a comment as well.

porcellus:
> In the enter OTP screen, we see "Resend Email" or "Resend SMS". But in the link sent screen, we see "Resend Link". For consistency, that should be "Resend Email" or "Resend SMS" as well.

I thought this was intentional since the design had Resend OTP and Resend Link if those were the only options and Resend Email/SMS if both was enabled.

porcellus:
> Instead, I should probably see no error. Perhaps the query param error should be removed as soon as it's read and displayed in the UI?

I tried for a fair bit of time today to make this work, but the problem is that the `SignInUp` component is re-created and loses state when we clear the query param meaning that the error message isn't actually displayed. I think this can be solved by further work, but it also feels like an issue we can deal with later if it causes confusion. Errors showing up again upon refreshes or when navigating back happens fairly often in my experience.

rishabhpoddar (approved):

rishabhpoddar:
# More UX feedback:
- [ ] If I seen an invalid OTP error, and then I click on the text input, the error goes away. So this is an issue if I click enter on the keyboard to see the error, then I focus the text input.
- [ ] Auto focus on the enter OTP box on page load
- [ ] Remove the placeholder in OTP input box
- [ ] Remove the placeholder in email / phone / email + phone input box
- [ ] Change default resend interval to 20 seconds

# In a separate PR
- [ ] Change the reset password screen to have a label for "Email" and remove the placeholder
- [ ] Remove ":" from all the labels

rishabhpoddar (approved):

rishabhpoddar (approved):

rishabhpoddar (approved):

rishabhpoddar (approved):

Issue #340, opened 2021-11-25
Style changes for ReactJS SDK


rishabhpoddar:
Center the social login buttons properly and use the social login provider's assets instead of making the button on our own

- [x] Center social buttons properly
- Use assets from the login provider for the social buttons instead of rendering our own - being done here: https://github.com/supertokens/supertokens-auth-react/issues/345
- [x] Hover state background color for buttons should be a darker version of the base color, currently it uses the SDK's default color (orange)

rishabhpoddar:
Released.

PR #344, opened 2021-12-21
Fix custom styles not working for forgot password and email verification. Fix centering of social provider buttons

nkshah2:
## Summary of change
- Fixes an issue where custom styles were not being passed to the forgot password and email verification components
- Fixes an issue where custom style for buttons would be overridden by default styling in hover and active states
- Fix styling for social login buttons:
    - Fix centering for the logo
    - Fix centering for the text

## Related issues

-  https://github.com/supertokens/supertokens-auth-react/issues/340

## Test Plan
No additional tests needed, tested with playground auth demo app for redwood js

## Documentation changes
None needed

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

None


lib/ts/recipe/thirdparty/components/themes/styles.ts
```diff
@@ -42,10 +42,12 @@ export function getStyles(palette: NormalisedPalette): NormalisedDefaultStyles {
         providerButton: {
             display: "flex",
             flexDirection: "row",
+            paddingLeft: "0px",
+            paddingRight: "0px",
         },
 
         providerButtonLeft: {
-            width: "34px",
+            width: "40px",
```
rishabhpoddar, line 50: Is hardcoding to another value the right solution for sure? How have you tested this?

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "supertokens-auth-react",
-    "version": "0.17.6",
+    "version": "0.17.7",
```
rishabhpoddar, line 3: You also need to edit package-lock. Did you install pre commit hook?

rishabhpoddar (changes requested):

lib/ts/recipe/thirdparty/components/themes/styles.ts
nkshah2: So the problem was that all buttons by default have a left padding of 6px, because of this the image for the provider would always seem off center even though it wasnt. Ive removed the left padding and added it to the width to keeo the layout consistent to how it appears now.

I tested it against the playground-auth app since I already had it set up

package.json
nkshah2: Done

lib/ts/recipe/thirdparty/components/themes/styles.ts
rishabhpoddar: Discuss on call..

PR #365, opened 2022-01-22
fix: multiple children were provided error

xuatz:
## Summary of change

There are 2 elements provided as the children of EmailPassword.EmailPasswordAuth, giving the error:

> This JSX tag's 'children' prop expects a single child of type 'Element', but multiple children were provided

Fixed it by adding a fragment.

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

## Notes

It seems like a really trivial change. But if there is any other work that is CRITICAL to merge this PR let me know.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
+## [unreleased]
+
+### Changes
+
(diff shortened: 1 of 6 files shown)


rishabhpoddar:
Thanks for this PR. Could you also make a similar change to other examples?
- with-thirdparty
- with-thirdpartyemailpassword

xuatz:
btw have you seen https://github.com/conventional-changelog/standard-version ?
how about using this to manage updating the changelogs?

rishabhpoddar:
> btw have you seen https://github.com/conventional-changelog/standard-version ?

This does seem interesting. Perhaps another issue for this would help?

PR #373, opened 2022-01-26
ci: add `lint-pr-title.yml` workflow

xuatz:
## Summary of change

Added a github action/workflow that will validate if the PR title follows [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)

## Details

If the PR title does not comply, this helpful failure message will appear.

image

### Step 1 of X to using `standard-version`

This is step one of the transition towards eventually adopting `standard-version` to automate semantic version bumping and changelog updates. But even if `standard-version` is not used, `conventional-commits` is also a nice standard to adopt to make reading git history slightly more pleasant.

Before officially adopting `standard-version` tho, the following should also be true:
* all the other projects should also adopt `conventional-commits` have this workflow (potentially could be solved with a monorepo structure, this makes docker-compose.yml easier too perhaps)
* there should be an equivalent for `standard-version` for other packages in other languages
* there should be a a custom script prepared to monkey patch any other part of the repo that needs version bumping but is not handled by `standard-version`

## Documentation changes

Updated CONTRIBUTING.md to give a heads up about conventional commits

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

### DIFF
--- a/.github/workflows/lint-pr-title.yml
+++ b/.github/workflows/lint-pr-title.yml
@@ -1,0 +1,20 @@
+name: "Lint PR Title"
+
+on:
+    pull_request:
+        types:
+            - opened
+            - reopened
+            - edited
+            - synchronize
+
+jobs:
+    pr-title:
+        name: Lint PR title
+        runs-on: ubuntu-latest
+        steps:
+            - uses: amannn/action-semantic-pull-request@v3
+              env:
+                  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+              with:
+                  validateSingleCommit: true
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [unreleased]
 
+-   add workflow to verify if pr title follows conventional commits
+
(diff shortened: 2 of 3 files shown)


rishabhpoddar:
hey @xuatz thanks for this PR. I'll keep this open for now until a similar PR has been made for all other repos we have:
- [ ] supertokens-core
- [ ]  supertokens-plugin-interface
- [ ] supertokens-mysql-plugin
- [ ]  supertokens-postgresql-plugin
- [ ]  supertokens-mongodb-plugin
- [ ] supertokens-node
- [ ]  supertokens-python
- [ ]  supertokens-golang
- [ ]  supertokens-react-native
- [ ]  supertokens-website

porcellus:
We may also want to consider something like this: https://changie.dev/

xuatz:
> We may also want to consider something like this: https://changie.dev/

Looks good! This solves the problem of looking for `standard-version` equivalent in other languages! 

Btw thinking about handling changelog in sibling repos, does supertokens have plans to one day move to a monorepo structure?

rishabhpoddar:
> Btw while talking about releasing changelogs in sibling repos, i was thinking, does supertokens have plans to one day move to a monorepo structure?

I don't think so. The dev process for different languages can be quite different.

Issue #233, opened 2021-04-19
Translatable components


fantoine:
Hi here,

I was wondering if there is a way to translate the components ?

One way could be to let us override some components to change the text, or just give us access to some keys that we could override.

rishabhpoddar:
Great idea! Will figure out the best way to enable this feature. 

Until then, you can modify the UI by overriding it the following way:

- Disabling the default UI for the sign in and all other pages by setting `disableDefaultImplementation` to true in `signInAndUpFeature` (and other features) when calling the `init` function (see https://supertokens.io/docs/auth-react/emailpassword/init). This will basically remove the UI from `/auth/*`. This mean you can handle this route to render anything you like.

- Then you would need to create your own component to show the login / reset pass etc UIs. For example, the login screen:

```js
import {SignInAndUp} from "supertokens-auth-react/recipe/emailpassword";

render() {
  return (
    <SignInAndUp>
      <YOUR UI COMPONENT GOES HERE>
    </SignInAndUp>
  );
}
```
- The default UI component can be found [in this source file](https://github.com/supertokens/supertokens-auth-react/blob/master/lib/ts/recipe/emailpassword/components/themes/signInAndUp/index.tsx) (for emailpassword recipe). So please feel free to copy this and modify the text to be in the language you like.

rishabhpoddar:
How do you imagine we could add this feature of translations? Any ideas please?

fantoine:
Thanks for the tip to translate components !

For translations, I guess you could add some `translation` object  to each recipes which need some texts.
It's also possible to integrate directly some libs like react-i18next but it may not be suitable for all projects.

What do you think about it ?

fantoine:
Have you also explored the possibility to use a React context instead of the `SuperTokens.init` singleton to share informations between components ?

rishabhpoddar:
Thanks for recommending `react-i18next`. I will check it out.

We wanted to use `SuperTokens.init` singleton because:
- The docs for it would be similar to that of initing supertokens on the backend (nodejs), or any other non-react framework (like angular or vue). 
- A recipe is defined as "more than a component". For example, the session recipe adds interceptors to `axios` or `fetch` and that should ideally be added independently of loading any component in the app.

Are you facing any issues due to the way we have designed this aspect of the lib?

fantoine:
No, I don't have any specific issue right now with this.
I get your point of view and agree with you about keeping a cross-framework way to do this. :)

[...]

PR #375, opened 2022-01-29
feat: make thirdpartypasswordless translateable

porcellus:
## Summary of change

Enables providing translations for thirdpartypasswordless

## Related issues

-   #233 

## Test Plan

- Visual testing would be great for this.
- Some tests check copies as well.

## Documentation changes

TODO:

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

-   [ ] Change base branch (or remove tppwless related parts) 
  - I used tppwless as a base branch because I wanted to extract all translations in one go.
-   [ ] Resolve TODOs in code
- [ ] Add test: Send general error with a translation key and make sure that the value of the translation key is rendered (and also FIELD_ERROR in third party).
- [ ] Test how we work together with i18next
- [ ] How will the user override the `t()` function?
- [ ] How does this behave in server side rendering?


rishabhpoddar:
# Feedback
- [ ] Destination should be 0.18 and not thirdpartyemailpassword
- [ ] The size check if failing - need to change the limits

lib/ts/recipe/emailpassword/translations.ts
```diff
@@ -0,0 +1,45 @@
+import { defaultTranslationsCommon } from "../../components/translations";
```
rishabhpoddar, line 1: Maybe consider moving all the translations into the theme folder of each recipe.

examples/for-tests/src/App.js
```diff
@@ -746,6 +746,14 @@ function getPasswordlessConfigs({ disableDefaultImplementation }) {
             disableDefaultImplementation,
             style: theme.style,
         },
+        translations: {
```
rishabhpoddar, line 749: We should consider changing the name to `translationsForLanguages`

rishabhpoddar, line 749: We should move `translations` to supertokens.init and not to have a recipe level, and club all the translation related config into one object as opposed to keeping them all in the root level. You can call this `translationsForLanguages` or `languageTranslation`. The fields in the object are:
- `translations` (contains the variable key values for different diff langs
- `t` function override
- `defaultLanguage`
- `translationScope`

rishabhpoddar: If this name is changing, we should also change the name of the functions in the index.ts files

lib/ts/index.ts
```diff
@@ -36,6 +37,14 @@ export default class SuperTokensAPIWrapper {
         return SuperTokens.getRoutingComponent();
     }
 
+    static changeLanguage(language: string): void {
```
rishabhpoddar, line 40: Also export useTranslations cause the user might want to use it during override

lib/ts/components/translationContext.tsx
```diff
@@ -0,0 +1,75 @@
...
+    const translateFunc = useCallback<TranslationFunc>(
+        (key: string, params: Record<string, string> = {}) => {
+            if (userTranslateFunc !== undefined) {
+                return userTranslateFunc(key, params);
+            }
+            const res = translationStore[currentLanguage] && translationStore[currentLanguage][key];
+            const fallback = translationStore[defaultLanguage] && translationStore[defaultLanguage][key];
+
+            if (res === undefined) {
+                console.warn(`Missing translation ${key} from ${currentLanguage}`);
```
rishabhpoddar, line 43: Do these only if in dev mode?

lib/ts/translationHelpers.tsx
```diff
@@ -0,0 +1,62 @@
+// type TranslationContext = Record<string, Record<Record<string, string>>>;
```
rishabhpoddar, line 1: Not sure why this comment is here..

lib/ts/translationHelpers.tsx
```diff
@@ -0,0 +1,62 @@
...
+    }
+}
+
+const CURRENT_LANGUAGE_COOKIE_NAME = "sCurrLanguage";
+
+export function saveCurrentLanguage(language: string, cookieDomain: string): void {
+    setFrontendCookie(CURRENT_LANGUAGE_COOKIE_NAME, language, cookieDomain);
+}
+export function getCurrentLanguageFromCookie(): string | null {
+    return getCookieValue(CURRENT_LANGUAGE_COOKIE_NAME);
```
rishabhpoddar, line 61: Add a try  catch and if that fails, ignore it and return `null`

lib/ts/translationHelpers.tsx
```diff
@@ -0,0 +1,62 @@
...
+            event,
+            handlerList.filter((h) => h !== handler)
+        );
+    }
+}
+
+const CURRENT_LANGUAGE_COOKIE_NAME = "sCurrLanguage";
+
+export function saveCurrentLanguage(language: string, cookieDomain: string): void {
+    setFrontendCookie(CURRENT_LANGUAGE_COOKIE_NAME, language, cookieDomain);
```
rishabhpoddar, line 58: Add a try  catch and if that fails, ignore it and return `null`

lib/ts/utils.ts
```diff
@@ -274,7 +274,115 @@ export class Deferred<T> {
...
+        } else {
+            res[key] = obj2[key];
+        }
+    }
+
+    return res;
+}
+
+// TODO: Reuse from website repo?
+export function normaliseCookieScopeOrThrowError(cookieScope: string): string {
```
rishabhpoddar, line 298: Maybe import from supertokens-website instead.

rishabhpoddar (changes requested):

lib/ts/recipe/thirdparty/translations.ts
```diff
@@ -0,0 +1,17 @@
...
+        ...defaultTranslationsCommon.en,
+        ...defaultTranslationsEmailVerification.en,
+
+        THIRD_PARTY_SIGN_IN_AND_UP_HEADER_TITLE: "Sign Up / Sign In",
+
+        THIRD_PARTY_SIGN_IN_UP_FOOTER_START: "By continuing, you agree to our",
+        THIRD_PARTY_SIGN_IN_UP_FOOTER_TOS: "Terms of Service",
+        THIRD_PARTY_SIGN_IN_UP_FOOTER_AND: "and",
+        THIRD_PARTY_SIGN_IN_UP_FOOTER_PP: "Privacy Policy",
+        THIRD_PARTY_SIGN_IN_UP_FOOTER_END: "",
```
rishabhpoddar, line 15: Missing "continue with <provider>" text?

lib/ts/recipe/thirdparty/translations.ts
```diff
@@ -0,0 +1,17 @@
...
+export const defaultTranslationsThirdParty = {
+    en: {
+        ...defaultTranslationsCommon.en,
+        ...defaultTranslationsEmailVerification.en,
+
+        THIRD_PARTY_SIGN_IN_AND_UP_HEADER_TITLE: "Sign Up / Sign In",
+
+        THIRD_PARTY_SIGN_IN_UP_FOOTER_START: "By continuing, you agree to our",
+        THIRD_PARTY_SIGN_IN_UP_FOOTER_TOS: "Terms of Service",
+        THIRD_PARTY_SIGN_IN_UP_FOOTER_AND: "and",
```
rishabhpoddar, line 13: Need to add sapces around the and

rishabhpoddar (changes requested):

porcellus:
This was manually merged into the thirdpartypasswordless PR

PR #378, opened 2022-02-03
feat: enable component translations

porcellus:
## Summary of change

Enables providing translations for all copies.

## Related issues

-   #233 

## Test Plan

- Added some tests that check that the right copy is shown
- Manual testing

## Documentation changes

- [ ] Provide translations
- [ ] Change language
- [ ] How to load a translation later
- [ ] Default language is also the fallback
- [ ] translation scope
- [ ] redefine translation function (like to with i18next)
- [ ] Using useTranslation in component override

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

-   [x] Change base branch (or remove tppwless related parts) 
  - I used tppwless as a base branch because I wanted to extract all translations in one go.
-   [x] Resolve TODOs in code
- [x] Add test: Send general error with a translation key and make sure that the value of the translation key is rendered (and also FIELD_ERROR in third party).
- [x] Test how we work together with i18next
- [x] How will the user override the `t()` function?
- [x] How does this behave in server side rendering?
- [x] In the country dropdown do we want the country names to go through translation?
- [ ] Should we review copy strings for consistency in this PR?
- [x] Reuse `normaliseCookieScopeOrThrowError` from website repo:
  - I don't think we should: 
    - it's exported with a different name making it specific to session scopes
    - it doesn't belong in the package interface (and submodule imports can lead to things breaking when we don't expect them)
    - it's not that much code anyway, and it almost never changes
- [ ] Manual testing
- [ ] Form fields translations


package.json
```diff
@@ -104,15 +104,15 @@
...
+            "limit": "115kb"
         },
         {
             "path": "recipe/thirdparty/index.js",
             "limit": "100kb"
         },
         {
             "path": "recipe/emailpassword/index.js",
-            "limit": "100kb"
+            "limit": "105kb"
```
rishabhpoddar, line 115: Size checks are still failing.

lib/ts/utils.ts
```diff
@@ -274,7 +274,118 @@ export class Deferred<T> {
...
+            res[key] = mergeObjects(res[key], obj2[key]);
+        } else {
+            res[key] = obj2[key];
+        }
+    }
+
+    return res;
+}
+
+export function normaliseCookieScopeOrThrowError(cookieScope: string): string {
```
rishabhpoddar, line 297: I think you had to reuse this from the supertokens-website lib

lib/ts/recipe/thirdpartyemailpassword/components/themes/translations.ts
```diff
@@ -0,0 +1,12 @@
+import { defaultTranslationsEmailPassword } from "../../../emailpassword/components/themes/translations";
+import { defaultTranslationsThirdParty } from "../../../thirdparty/components/themes/translations";
+
+export const defaultTranslationsThirdPartyEmailPassword = {
+    en: {
+        ...defaultTranslationsThirdParty.en,
+        ...defaultTranslationsEmailPassword.en,
+
+        THIRD_PARTY_EMAIL_PASSWORD_SIGN_IN_AND_UP_HEADER_TITLE: "Sign Up or Log In",
+        THIRD_PARTY_EMAIL_PASSWORD_SIGN_IN_AND_UP_DIVIDER_OR: "or",
```
rishabhpoddar, line 10: You need a space around or?

rishabhpoddar (changes requested):

lib/ts/recipe/thirdpartyemailpassword/components/themes/translations.ts
porcellus: No, this doesn't have any text around it.

package.json
porcellus: I updated this (but the sizes didn't change so it will not add a new comment I think)

lib/ts/utils.ts
porcellus: I moved this (and all other TODOs in the code) into the list in the PR description. In short, I don't think we should because:
- We'd need to rename it in the website repo (right now, it's `normaliseSessionScopeOrThrowError`)
- Submodule imports like this (going into the utils of another package) are not nice
- It's relatively short and really stable code, so we are unlikely to have any trouble with updating it

(the first one is something I didn't realize the last time we spoke about this)

rishabhpoddar: Fair enough. Another idea to consider is moving this to the same repo in which you have the JS override stuff?

lib/ts/recipe/thirdpartyemailpassword/components/themes/translations.ts
rishabhpoddar: okay!

rishabhpoddar:
# Feedback
- [ ] If i remove `EMAIL_PASSWORD_EMAIL_LABEL` (which is a form field label) from the translation.ts file, `npm run build && npm run lint` doesn't complain about it. But if I do this for other ones which are non form field translation keys, it works as expected. Do you want to do anything about this? (I'm OK leaving it as it is)
- [ ] Size check is failing.
- [ ] Translation keys like `THIRD_PARTY_SIGN_IN_UP_FOOTER_START` (and other recipe ones) should have a space at the end of the sentence?
- [ ] In passwordless, the button for "Change email" in the enter OTP screen doesn't have a space on the left:
Screenshot 2022-02-10 at 11 38 47

porcellus:
> If i remove `EMAIL_PASSWORD_EMAIL_LABEL` ...
I'm not sure how well we can handle situations like those, maybe we could have "restricted" prop names that can only hold translation keys, but I'm not sure if it's worth it.
> `THIRD_PARTY_SIGN_IN_UP_FOOTER_START`
I don't think so, it looks good, the `a` tag seems to insert enough space before it. Maybe I could also remove them around the and, but It looks OK I think.

Issue #380, opened 2022-02-04
Name of overridable component in map is not the same as what is shown in component inspector


rishabhpoddar:
In ThirdPartyEmailPassword recipe, we see that the sign up form component is called `SignInAndUp` in the react inspector. But in the override map, it's called `EmailPasswordSignUpForm`.

We need to change it so that we see the right name in the inspector.

rishabhpoddar:
It has already been released in 0.19

PR #386, opened 2022-02-10
refactor!: rework feature components for easier composition

porcellus:
## Summary of change
Extracting logic of feature components into custom hooks

## Related issues

-   https://github.com/supertokens/supertokens-core/issues/330 this will make the react implementation easier/cleaner

## Test Plan

Should cause no change in behavior

## Documentation changes

Should cause no change in behavior

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

## TODO
- [x] Update internal contributing guide with new principles that are added
- [x] Update demo apps in this repo to use new version and any code changes that might be required for them (some of them use JS to manipulate the react UI, so that might need to change)
- [x] Since this is a breaking change, we should also solve this bug in this release: https://github.com/supertokens/supertokens-auth-react/issues/380
- [x] Component override for ThirdPartyForm was showing as Anonymous.
- [x] Update documentation about overriding
- [ ] Update all demo apps


rishabhpoddar:
It's really hard to review this on a code level. I'd appreciate comments in the code highlighting how the new structure works on a high level. If not in the code, please add to the contributing docs in the relevant place. Thanks

porcellus:
I'll write a quick summary here first and then work on adding them to the contributions doc, I think this belongs there instead of in one of the recipes.
So the main changes:
- `GeneralError` moved out of `FormBase` (possibly breaking interface change for overrides)
- Feature component logic is extracted to custom hooks we can reuse for easier composition
- Using extracted hooks for `thirdpartyemailpassword` instead of embedding through the feature comps

lib/ts/recipe/emailpassword/components/features/signInAndUp/index.tsx
```diff
@@ -19,228 +19,276 @@
...
-        this.state = {
+export const useFeatureReducer = (recipe: Recipe | undefined) => {
+    return React.useReducer(
+        (oldState: SignInAndUpState, action: EmailPasswordSignInAndUpAction) => {
+            switch (action.type) {
+                case "setSignIn":
+                    return {
+                        ...oldState,
+                        isSignUp: false,
+                    };
```
rishabhpoddar, line 49: is this missing a break;?

lib/ts/recipe/emailpassword/types.ts
```diff
@@ -425,13 +390,13 @@ export type FormBaseProps<T> = {
...
-    error?: string;
-
     validateOnBlur?: boolean;
 
+    clearError: () => void;
+    onError: (error: string) => void;
     onSuccess?: (result: T & { status: "OK" }) => void;
 
-    callAPI: (fields: APIFormField[]) => Promise<FormBaseAPIResponse<T>>;
+    callAPI: (fields: APIFormField[], setValue: (id: string, value: string) => void) => Promise<FormBaseAPIResponse<T>>;
```
rishabhpoddar, line 399: What is the purpose of `setValue` function?

rishabhpoddar (changes requested):

lib/ts/recipe/emailpassword/types.ts
porcellus: It's used in `emailOrPhoneForm`: we can update the stored value after submitting i.e., guess the international number.

lib/ts/recipe/emailpassword/components/features/signInAndUp/index.tsx
porcellus: The return statements return from the entire function, making breaks not necessary.

porcellus:
#380 should be solved since the referenced component was removed.

rishabhpoddar:
# Feedback
- [ ] Title and sub title missing in reset password forms (both the forms)

PR #398, opened 2022-03-08
feat: Replace supertokens-website for session recipe with supertokens-web-js

nkshah2:
## Summary of change

- Removes dependency on supertokens-website
- Refactors Session recipe to use supertokens-web-js

## Related issues

## Test Plan

All existing test cases pass (except for passwordless, which is expected)
Screenshot 2022-03-08 at 1 20 09 PM

## Documentation changes

WIP

    -   ~~In `package.json`~~
    -   ~~In `lib/ts/version.ts`~~
    -   ~~To know which one it is, run find the latest released tag (`git tag`) in the format `vX.Y.Z`, and then find the latest branch (`git branch --all`) whose `X.Y` is greater than the latest released tag.~~
    -   ~~If no such branch exists, then create one from the latest released branch.~~

-   [ ] Verify Size Limit


lib/ts/recipe/session/recipe.ts
```diff
@@ -116,8 +118,12 @@ export default class Session extends RecipeModule<unknown, unknown, unknown, any
     private async getSessionContext({ action }: RecipeEvent): Promise<SessionContextType> {
```
rishabhpoddar, line 118: I think this can take a userContext in it since this is called in SessionAuth, and that should also take a userContext in it.

rishabhpoddar (changes requested):

lib/ts/recipe/session/recipe.ts
rishabhpoddar: This is still not done.

rishabhpoddar (changes requested):

lib/ts/recipe/session/recipe.ts
nkshah2: Done

rishabhpoddar (approved):

PR #400, opened 2022-03-19
feat(sessiongrants): add support for session grants

porcellus:
## Summary of change

Add support for session grants

## Related issues

- supertokens/core-driver-interface#36
- supertokens/supertokens-core#390
- supertokens/supertokens-node#278

## Test Plan

TODO

## Documentation changes

- [ ] All examples in the docs for adding claimValidators to SessionAuth, we should have useMemo.
- [ ] Need to show how to modify / override the default claims

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

-   [ ] Check interface


lib/ts/recipe/session/recipe.ts
```diff
@@ -114,7 +114,7 @@ export default class Session extends RecipeModule<unknown, unknown, unknown, any
     };
 
     private async getSessionContext({ action }: RecipeEvent): Promise<SessionContextType> {
-        if (action === "SESSION_CREATED" || action === "REFRESH_SESSION") {
+        if (action === "SESSION_CREATED" || action === "REFRESH_SESSION" || action === "GRANT_MISSING") {
```
rishabhpoddar, line 117: We are checking if the action type if `GRANT_MISSING`, but not doing anything specific for it in the if condition?

lib/ts/recipe/session/sessionAuth.tsx
```diff
@@ -35,7 +35,9 @@ type PropsWithAuth = {
 };
 
 type Props = (PropsWithoutAuth | PropsWithAuth) & {
+    requiredGrants?: Grant<any>[];
     onSessionExpired?: () => void;
+    onMissingGrant?: (grantKey: string) => void;
```
rishabhpoddar, line 40: Should we call it `grantId`? Cause we use that everywhere, for example, in the header response we use `missing-grant-id`

rishabhpoddar (changes requested):

rishabhpoddar:
Should we make changes to the supertokens-website SDK first before making changes here?

lib/ts/recipe/session/recipe.ts
porcellus: The session context is the same in that case, we still have a userId and the payload.

porcellus:
We should definitely make changes to `supertokens-website` first when implementing, this PR is more about the interface we are planning.

rishabhpoddar:
# Feedback (WIP):
- [ ] We should rename event “API_MISSING_CLAIM” to “MISSING_CLAIM” and add a isFromAPI boolean to event metadata.
- [ ] We should rename “ACCESS_TOKEN_UPDATED” to “ACCESS_TOKEN_PAYLOAD_UPDATED”
- [ ] We should rename `onMissingClaim` to `handleMissingClaim`
- [ ] If `handleMissingClaim` is `undefined`, then auth wrapper should fire event, else it should call the function. If the function returns `true`, it should not fire event, else it should fire event.
- [ ] Make role keys in access token smaller -> `st-email-verification` -> `st-ev`
- [ ] User context passing on the frontend.

rishabhpoddar:
Instead of adding an elaborate API right at the start, what if we do the following:
- Ask users to handle missing claims for every API call they make. We can provide helper functions to extract the missing claims from the API calls like:
   ```
   try {
      await axios(...);
   } catch (e) {
      let missingClaim = supertokens.getMissingClaim(e); // this should also take a response from `fetch`
      if (missingClaim !== undefined) {
         // handle missing claim yourself... If you don't handle this, you can also treat this as a generic Something went wrong error as in most cases, it shouldn't come here if your app flow is setup correctly, and if the user is not purposely trying to manipulate things.
      }
   }
   ```
   - One question here is for things like missing role, how will they know which role is required and which is missing?
   - What is the structure of `missingClaim`?

- Instead of making SessionAuth take a requiredClaim array, we can ask users to check for required claims in their component. This is simpler API for us to explain and document + for them to be able to handle:
   ```
   let {accessTokenPayload} = useSessionContext();
   if (supertokens.hasClaimInAccessTokenPayload(accessTokenPayload, RoleClaim.hasRole("admin")) {
     // TODO..
   } else {
      // TODO..
   } 
   ```

- For things like Email verification where we want to handle it automatically, we can make the SessionAuth wrapper check the `accessTokenPayload` for if email verification claim is present or not. If it isn't we will navigate to the email verification page. Sure, we will miss cases of if their API throws missing email verification error, but that should never happen if they want the user to use their app only if their email is verified (cause the SessionAuth would always redirect).
- We do not need to fire events for missing claims in supertokens-website repo as the above API would be good enough.

-----------
One problem with the above is that there are two new functions:
- `getMissingClaim`
- `hasClaimInAccessTokenPayload`

Do you think it's possible to combine them into one function? What would such a function be called?

porcellus:
1. I think we'd need to do this whatever API we provide. The requests should return an error (or throw) anyway; adding a helper function to get the id of the missing claim is a nice extra thing.
2. I think this might lead to code repetition, but it could work out with good recommendations. I think this provides a bit more control/granularity at the cost of implementing things in the app instead of the SDK.
3. `getMissingClaim` and `hasClaimInAccessTokenPayload` have different params, so I don't think it would make sense to combine them. 
  - I'd rename `hasClaimInAccessTokenPayload` to `checkClaims` (and have it on the session recipe) because it may call their backend during checking.
  - I'd rename `getMissingClaim` to `getMissingClaimFromResponse` (or Error)

porcellus:
1. Structure of missing claims returned by the backend:
  - We should clarify that the id is the id of the claim validator (and maybe rename the id of the claim class to key)
  - We should have a metadata obj in the invalid claim response header (and in the body)
  - Change the `isValid` function of claim checkers (-> validators):
    - Rename to `validate`
    - It should return an object with the shape: 
    ```ts 
    type ClaimValidationResult = {isValid: true} || {isValid: false, reason?: JSONObject}
    ```
    - rename to ClaimValidator
 2. Session context:
   - should provide `invalidClaim?: { id: string; reason?: JSONObject}`
   - remove `handleMissingClaim`
 3. Session recipe:
   - provide `async validateClaims(...)`
   - fire `API_INVALID_CLAIM` event
 4. Figure out a way to add default handlers (and claim validators) from other recipes (e.g., EmailVerification)
 5. Renaming:
   - claimCheckers
 7. Remove:
   - `defaultClaimValidators`

examples/with-claims/api/claims/MFAClaim.ts
```diff
@@ -0,0 +1,37 @@
...
+                SecondFactorOTPClaim.validators.hasFreshValue(true, maxAgeInSeconds).validate(payload, ctx) ||
+                SecondFactorTOTPClaim.validators.hasFreshValue(true, maxAgeInSeconds).validate(payload, ctx),
+        };
+    },
+    any3FreshFactors(maxAgeInSeconds: number) {
+        return {
+            validatorTypeId: "st-mfa",
+            validate: (payload, ctx) =>
+                SecondFactorOTPClaim.validators.hasFreshValue(true, maxAgeInSeconds).validate(payload, ctx) &&
+                SecondFactorTOTPClaim.validators.hasFreshValue(true, maxAgeInSeconds).validate(payload, ctx),
```
rishabhpoddar, line 34: How is this any 3 fresh factors?

examples/with-claims/api/claims/MFAClaim.ts
```diff
@@ -0,0 +1,37 @@
+import { BooleanClaim } from "supertokens-node/lib/build/recipe/session/claimBaseClasses/booleanClaim";
+
+// All of these claims/checkers are exported by us and added by endpoints of an MFA recipe
+export const SecondFactorOTPClaim = new BooleanClaim({
+    fetch: () => undefined,
```
rishabhpoddar, line 5: Why are we returning `undefined` here? Should we return `false`?

examples/with-claims/api/claims/MFAClaim.ts
```diff
@@ -0,0 +1,37 @@
...
+    key: "st-mfa-otp",
+});
+
+export const SecondFactorTOTPClaim = new BooleanClaim({
+    fetch: () => undefined,
```
rishabhpoddar, line 10: Why are we returning `undefined` here? Should we return `false`?

examples/with-claims/api/claims/RolesClaim.ts
```diff
@@ -0,0 +1,91 @@
...
+            including: (role: string) => ({
+                validatorTypeId: this.key,
+                claim: this,
+                shouldRefetch: (payload, ctx) => {
+                    return this.getValueFromPayload(payload, ctx) === undefined;
+                },
+                validate: async (payload, ctx) => {
+                    return {
+                        isValid: this.getValueFromPayload(payload, ctx).includes(role),
+                        reason: { missingRole: role },
```
rishabhpoddar, line 18: Why do we give a reason if `isValid` is `true`?

examples/with-claims/api/claims/RolesClaim.ts
```diff
@@ -0,0 +1,91 @@
+import Session, { SessionClaim, SessionClaimValidator, SessionContainer } from "supertokens-node/recipe/session";
+import { roleDB } from "../mockRoleDb";
+
+class RolesClaimClass extends SessionClaim<string[]> {
+    constructor() {
+        super("st-r");
+        this.hasRole = {
+            validatorTypeId: this.key,
+            including: (role: string) => ({
```
rishabhpoddar, line 9: I don't think we should `including`. Just `RolesClaimClass.hasRole([".."]);`

examples/with-claims/api/claims/RolesClaim.ts
```diff
@@ -0,0 +1,91 @@
...
+        delete res[this.key];
+        return res;
+    }
+
+    getValueFromPayload(payload: any, _userContext: any): string[] {
+        return payload[this.key]?.v;
+    }
+
+    fetch(userId: string, ctx: any) {
+        return roleDB.get(userId) || [];
```
rishabhpoddar, line 87: Shoudl be updated to use our user roles recipe. 

What if the user has their own way to store roles? How will they override this to get that value?

examples/with-claims/api/claims/RolesClaim.ts
```diff
@@ -0,0 +1,91 @@
...
+        const origRoles = this.getValueFromPayload(payload, userContext);
+        const newRoles = Array.from(new Set([...origRoles, role]));
+        return session.mergeIntoAccessTokenPayload(this.addToPayload_internal({}, newRoles, userContext));
+    }
+
+    setRoles(session: SessionContainer, roles: string[], userContext?: any) {
+        return session.mergeIntoAccessTokenPayload(this.addToPayload_internal({}, roles, userContext));
+    }
+
+    updateRoles(session: SessionContainer, userContext?: any) {
```
rishabhpoddar, line 40: Consider calling this `refreshRolesFromDb` or something. It's very confusing with `addRole`

examples/with-claims/api/claims/RolesClaim.ts
```diff
@@ -0,0 +1,91 @@
...
+                    };
+                },
+            }),
+        };
+    }
+    readonly hasRole: {
+        validatorTypeId: string;
+        including: (role: string) => SessionClaimValidator;
+    };
+
```
rishabhpoddar, line 28: - missing removeRole function
- missing get all roles function

examples/with-claims/api/claims/RolesClaim.ts
```diff
@@ -0,0 +1,91 @@
...
+    addRole(session: SessionContainer, role: string, userContext?: any) {
+        const payload = session.getAccessTokenPayload(userContext);
+        const origRoles = this.getValueFromPayload(payload, userContext);
+        const newRoles = Array.from(new Set([...origRoles, role]));
+        return session.mergeIntoAccessTokenPayload(this.addToPayload_internal({}, newRoles, userContext));
+    }
+
+    setRoles(session: SessionContainer, roles: string[], userContext?: any) {
+        return session.mergeIntoAccessTokenPayload(this.addToPayload_internal({}, roles, userContext));
+    }
```
rishabhpoddar, line 38: Don't need this IMO

examples/with-claims/api/claims/RolesClaim.ts
```diff
@@ -0,0 +1,91 @@
+import Session, { SessionClaim, SessionClaimValidator, SessionContainer } from "supertokens-node/recipe/session";
```
rishabhpoddar, line 1: This should be tightly couple with our user roles recipe. For example, the developer should not have to call `addRole` on this claim AND in the session recipe.

Also, this is missing concept of permissions

examples/with-claims/api/claims/RolesClaim.ts
```diff
@@ -0,0 +1,91 @@
...
+import { roleDB } from "../mockRoleDb";
+
+class RolesClaimClass extends SessionClaim<string[]> {
```
rishabhpoddar, line 4: Can't we just use extends PrimitiveClaim?

examples/with-claims/api/index.ts
```diff
@@ -0,0 +1,197 @@
...
+                                roleDB.set(res.user.id, ["user"]);
+                            }
+                        }
+                        return res;
+                    },
+                }),
+            },
+        }),
+        Session.init({
+            claimsToAddOnCreation: [RolesClaim, EmailVerifiedClaim],
```
rishabhpoddar, line 53: Also show how we can add custom claims (simple key value)

examples/with-claims/api/index.ts
```diff
@@ -0,0 +1,197 @@
...
+
+app.use(middleware());
+
+app.get(
+    "/api1",
+    verifySession({
+        overwriteDefaultValidators: [
+            RolesClaim.hasRole.including("admin"),
+            EmailVerifiedClaim.isVerified, // Maybe this could be just EmailVerifiedClaim
+            SecondFactorCheckers.any2Factors,
```
rishabhpoddar, line 85: It's weird that the others are called `*Claims` and this is called `*Checkers`

examples/with-claims/api/index.ts
```diff
@@ -0,0 +1,197 @@
...
+        ],
```
rishabhpoddar, line 86: Uhmm. It's a little strange that we need to specify `EmailVerifiedClaim.isVerified` again. I know it's under `overwriteDefaultValidators`, but maybe we should instead have something like:
```
verifySession({
   getClaimsToVerify: (defaultClaims) => {
      return [...defaultClaims, RolesClaim.hasRole.including("admin"), SecondFactorCheckers.any2Factors]
   }
})

```

examples/with-claims/api/index.ts
```diff
@@ -0,0 +1,197 @@
...
+    }),
+    async (req, res) => {
+        const session: SessionContainer = (req as any).session;
+
+        const roles = RolesClaim.getValueFromPayload(session.getAccessTokenPayload(), {});
```
rishabhpoddar, line 91: This is still quite complex of an API. Something like:
```
RolesClaim.getRolesFromSession(session)
```

is better

examples/with-claims/api/index.ts
```diff
@@ -0,0 +1,197 @@
...
+
+        res.send({
+            status: "OK",
+        });
+    }
+);
+
+app.post(
+    "/refresh-email-verified",
+    verifySession({ overwriteDefaultValidators: [] }), // This only requires the default EmailVerifiedClaim
```
rishabhpoddar, line 138: This is misleading that even if you are providing an empty array, we are still applying email verification claim?

examples/with-claims/api/index.ts
```diff
@@ -0,0 +1,197 @@
...
+        await RolesClaim.updateRoles(session);
+
+        res.send({
+            status: "OK",
+        });
+    }
+);
+
+app.post(
+    "/refresh-email-verified",
```
rishabhpoddar, line 137: Shouldn't this be more like `/set-email-verification-to-true`?

examples/with-claims/src/App.js
```diff
@@ -0,0 +1,149 @@
...
+SuperTokens.init({
+    appInfo: {
+        appName: "SuperTokens Demo App", // TODO: Your app name
+        apiDomain: getApiDomain(), // TODO: Change to your app's API domain
+        websiteDomain: getWebsiteDomain(), // TODO: Change to your app's website domain
+    },
+    recipeList: [
+        EmailPassword.init({
+            emailVerificationFeature: {
+                mode: "OFF",
```
rishabhpoddar, line 40: It's strange that this is given "OFF", but the backend still has email verification as a require default claim.

examples/with-claims/src/MissingClaimPopup/index.jsx
```diff
@@ -0,0 +1,17 @@
+import { useEffect } from "react";
+import { useNavigate, useLocation } from "react-router-dom";
+
+export default function MissingClaimPopup({ missingClaim }) {
```
rishabhpoddar, line 4: Where is this being used?

examples/with-claims/src/routes/common.component.jsx
```diff
@@ -0,0 +1,145 @@
...
+    const api = useTestApi();
+
+    const logs = useTestLogs();
+
+    const session = useSessionContext();
+
+    if (session.invalidClaim) {
+        if (session.invalidClaim.validatorId === RolesClaim.hasRole.id) {
+            // Update rolesclaim like this
+            RolesClaim.hasRole.withValue("admin");
```
rishabhpoddar, line 25: This is not a very intuitive API.

rishabhpoddar (changes requested):

examples/with-claims/src/routes/common.component.jsx
porcellus: This should've been removed (or it should be `await RolesClaim.refresh()`) this was originally demonstrating something different and since it's a no-op I forgot it here.

examples/with-claims/src/MissingClaimPopup/index.jsx
porcellus: I didn't end up using this, I'll remove it.

examples/with-claims/api/index.ts
porcellus: This should actually refresh it from the DB instead of setting it to true, I'll update it.

examples/with-claims/api/claims/MFAClaim.ts
porcellus: That was added by accident

examples/with-claims/api/claims/MFAClaim.ts
porcellus: We don't necessarily want the false info in the access token: it's not useful (we are always checking for true value) and it grows the token. Also, this was intended as a demonstration for claims only added manually in other endpoints instead of on session creation.
In an implementation where the 2fa is optional and/or the user can configure what factors they want to be enabled then we could consider adding the factor set to false as it actually contains information. (This is the same basic idea as adding a state for each factor into the token)

porcellus: Still, I don't think we'd need that info in the token.

examples/with-claims/api/claims/RolesClaim.ts
porcellus: That's not necessary (and incorrectly typed), but it doesn't cause any errors. I'll update it.

examples/with-claims/api/claims/RolesClaim.ts
porcellus: This is for symmetry with the front-end, where we may want to use `hasRole.validatorTypeId`

examples/with-claims/api/claims/RolesClaim.ts
porcellus: We could, but it'd be slightly wrong: this contains an array, not a primitive value.

examples/with-claims/api/claims/RolesClaim.ts
porcellus: It does, but I'm not sure it's worth including in this example. When we provide a full claims implementation in the roles recipe it will export those claims instead of the user (or this example) implementing them.

examples/with-claims/api/claims/MFAClaim.ts
rishabhpoddar: Alright!

examples/with-claims/api/claims/RolesClaim.ts
rishabhpoddar: `RoleClaim.hasRoles.includes("admin")`
`RoleClaim.hasRoles.includesAll(["admin", "..."])`
`RoleClaim.hasRoles.includesAny(["admin", "..."])`
`RoleClaim.getRolesFromPayload(accessTokenPayload)`

(Also, these should go in supertokens-web-js?)

examples/with-claims/api/claims/RolesClaim.ts
rishabhpoddar: Let's discuss this part later

examples/with-claims/api/claims/RolesClaim.ts
rishabhpoddar: Im going to keep this open so that we don't forget about supporting permissions.

rishabhpoddar: By default, other auth providers add the user's roles and permissions into the JWT.

examples/with-claims/src/App.js
rishabhpoddar: You shouldn't have to set the email verification mode at all on the frontend anymore.

examples/with-claims/api/claims/EmailVerifiedClaim.ts
```diff
@@ -0,0 +1,30 @@
+import { SessionClaimValidator } from "supertokens-node/recipe/session";
+import { BooleanClaim } from "supertokens-node/lib/build/recipe/session/claimBaseClasses/booleanClaim";
+import { isEmailVerified } from "supertokens-node/recipe/emailpassword";
+
```
rishabhpoddar, line 4: If this is claim is not met, we want to automatically navigate the user to the mail verification screen - the user should not have to manually check sessionContext and navigate there.

This is unlike other claims (like Roles) where the user has to decide what to do in case a role is not met

rishabhpoddar (changes requested):

examples/with-claims/src/App.js
```diff
@@ -0,0 +1,148 @@
...
+    // 2. the api call adds a log
+    // 3. adding the log rerenders this component
+    // 4. the rerender changes the references of the claims which causes the checking to restart
+    // 5. since the check (and the refresh) didn't complete the first time, it calls refresh again...
+    // This is not normally the case, the claim refreshing shouldn't rerender the auth component,
+    // it's just a weird test setup because of logging.
+    const route1Claims = useMemo(
+        () => [EmailVerifiedClaim.isVerified, RolesClaim.hasRole.including("admin"), MFAClaim.completed2FA()],
+        []
+    );
```
rishabhpoddar, line 87: We can change this to using a function that takes a DefaultClaims array.

rishabhpoddar (changes requested):

examples/with-claims/src/App.js
rishabhpoddar: Different modes:
- REQUIRED: Means that the default claims would have email verification automatically and redirect
- OPTIONAL: Means that the default claims would have email verification automatically and NOT redirect
- OFF: Means that default claim doesn't have email verification automatically. If the user puts it in the claims themselves for a route, then it should check it and NOT redirect.


One another interpretation could be:
- REQUIRED: Means that the default claims would have email verification automatically and redirect
- OPTIONAL: Means that the default claims would NOT have email verification automatically, but if the user provides it for a route, it SHOULD redirect
- OFF: Means that default claim doesn't have email verification automatically. If the user puts it in the claims themselves for a route, then it should check it and NOT redirect.
----------------------
OR, we should have two config values: `checkForAllRoutes` and `redirectAutomatically`.

rishabhpoddar (changes requested):

lib/ts/recipe/session/sessionAuth.tsx
```diff
@@ -109,28 +168,35 @@ const SessionAuth: React.FC<PropsWithChildren<Props>> = ({ children, ...props })
...
-                    setContext(event.sessionContext);
+                    setContextUpdate(event.sessionContext);
                     return;
                 case "ACCESS_TOKEN_PAYLOAD_UPDATED":
-                    setContext(event.sessionContext);
+                    setContextUpdate(event.sessionContext);
+                    return;
+                case "API_INVALID_CLAIM":
+                    // In general the user should not be calling APIs that throw this
+                    // so this may suggest that a claim was invalidated
```
rishabhpoddar, line 181: Why not?

examples/with-claims/api/index.ts
```diff
@@ -0,0 +1,201 @@
...
+app.use(morgan("dev") as any);
+app.use(bodyParser.json());
+
+app.use(middleware());
+
+app.get(
+    "/api1",
+    verifySession({
+        overwriteDefaultValidators: [
+            RolesClaim.hasRole.including("admin"),
```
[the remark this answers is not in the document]
porcellus: It already does, the reason contains the missing role.

examples/with-claims/api/index.ts
```diff
@@ -0,0 +1,201 @@
...
+            EmailVerifiedClaim.isVerified, // Maybe this could be just EmailVerifiedClaim
+            SecondFactorClaimValidators.any2Factors,
```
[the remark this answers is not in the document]
porcellus: We could add something like "claim combinators", so functions like `and` and `or`, but I don't see a practical way of chaining with `&&`.
I think we could cover almost all use-cases by providing functions like `includesOneOf`, `includesAll` etc on each claim. I think this feature can launch without those combinators since it's solvable by overrides and/or custom claims. If this is a recurring thing we can implement those later.

examples/with-claims/src/App.js
```diff
@@ -0,0 +1,148 @@
...
+    // These have to be memoized in this test, because:
+    // 1. the refresh of a claim calls an api
+    // 2. the api call adds a log
+    // 3. adding the log rerenders this component
+    // 4. the rerender changes the references of the claims which causes the checking to restart
+    // 5. since the check (and the refresh) didn't complete the first time, it calls refresh again...
+    // This is not normally the case, the claim refreshing shouldn't rerender the auth component,
+    // it's just a weird test setup because of logging.
+    const route1ClaimOverride = useCallback(
+        (defaultClaims) => [...defaultClaims, RolesClaim.hasRole.includes("admin"), MFAClaim.completed2FA()],
```
rishabhpoddar, line 84: should use globalClaims instead of defaultClaims?

lib/ts/recipe/session/types.ts
```diff
@@ -23,9 +24,13 @@ export type RecipeEvent =
...
           action: "UNAUTHORISED";
           sessionExpiredOrRevoked: boolean;
+      }
+    | {
+          action: "API_INVALID_CLAIM";
+          claimId: string;
       };
 
-export type RecipeEventWithSessionContext = RecipeEvent & { sessionContext: SessionContextType };
+export type RecipeEventWithSessionContext = RecipeEvent & { sessionContext: SessionContextTypeWithoutInvalidClaim };
```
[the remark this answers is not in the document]
rishabhpoddar: > The session context contains the result of claim checking, while the event sent here can't

I don't understand this. What are the implications of this from a user's point of view?

examples/with-claims/api/index.ts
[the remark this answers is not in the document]
rishabhpoddar: cool

lib/ts/recipe/session/recipe.ts
```diff
@@ -93,6 +101,55 @@ export default class Session extends RecipeModule<unknown, unknown, unknown, any
...
+        return undefined;
+    };
+
+    addDefaultClaimValidator = (claimValidator: SessionClaimValidator<any>) => {
+        this.defaultClaimValidators.push(claimValidator);
+    };
+
+    getDefaultClaimValidators = () => {
+        return this.defaultClaimValidators;
+    };
```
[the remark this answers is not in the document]
rishabhpoddar: Yea.. i think we should, just like how it's done on the backend (via overriding of `getGlobalClaimValidators` function).

rishabhpoddar (changes requested):

lib/build/recipe/session/sessionAuth.js
```diff
@@ -312,4 +432,9 @@ var SessionAuth = function (_a) {
     }
     return react_1.default.createElement(sessionContext_1.default.Provider, { value: context }, children);
 };
+// const SessionAuthWithKey: React.FC<Props> = (props) => (
+//     <SessionAuth key={`st-${props.requireAuth}-${props.claimValidators?.map((p) => p.id)}`} {...props}>
+//         {props.children}
+//     </SessionAuth>
+// );
```
bikkimahato, line 439: Commented code should not be pushed to GitHub.

examples/with-claims/api/index.ts
```diff
@@ -0,0 +1,219 @@
...
+    }
+);
+
+app.post(
+    "/refresh-email-verified",
+    verifySession({ overrideGlobalClaimValidators: () => [] }) as any, // This requires no claims
+    async (req, res) => {
+        const session: SessionContainer = (req as any).session;
+
+        await session.fetchAndGetAccessTokenPayloadUpdate(EmailVerifiedClaim);
```
rishabhpoddar, line 164: Shouldn't this be fetchAndSetClaim?

examples/with-claims/src/test.context.jsx
```diff
@@ -0,0 +1,124 @@
+import { useState } from "react";
```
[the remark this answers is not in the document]
rishabhpoddar: @porcellus missed this comment?

lib/ts/recipe/session/index.ts
```diff
@@ -53,6 +53,19 @@ export default class SessionAPIWrapper {
...
+    }
+
+    // have backwards compatibility to allow input as "signin" | "signup"
+    static redirectToAuth(input?: { redirectBack?: boolean }): Promise<void> {
+        if (input?.redirectBack !== true) {
+            return Session.getInstanceOrThrow().redirectToAuthWithoutRedirectToPath();
+        } else {
+            return Session.getInstanceOrThrow().redirectToAuthWithRedirectToPath();
+        }
+    }
```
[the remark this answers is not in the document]
rishabhpoddar: @porcellus missed this?

lib/ts/recipe/session/recipe.ts
```diff
@@ -93,6 +101,55 @@ export default class Session extends RecipeModule<unknown, unknown, unknown, any
...
+        }
+        queryParams = {
+            ...queryParams,
+            redirectToPath,
+        };
+        return this.redirectToAuthWithoutRedirectToPath(history, queryParams);
+    }
+
+    redirectToAuthWithoutRedirectToPath(history?: any, queryParams?: Record<string, string>) {
+        const redirectUrl = appendQueryParamsToURL(this.config.appInfo.websiteBasePath, queryParams);
```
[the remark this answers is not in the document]
rishabhpoddar: @porcellus missed this?

lib/ts/recipe/session/types.ts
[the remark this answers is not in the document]
rishabhpoddar: @porcellus missed this?

lib/ts/recipe/session/sessionAuth.tsx
```diff
@@ -55,12 +55,65 @@ const SessionAuth: React.FC<PropsWithChildren<Props>> = ({ children, ...props })
...
+                    if (await validator.shouldRefresh(accessTokenPayload, userContext)) {
+                        await validator.refresh(userContext);
+                        accessTokenPayload = await session.current.getAccessTokenPayloadSecurely();
+                    }
+                    const validationRes = await validator.validate(accessTokenPayload, userContext);
+                    if (!validationRes.isValid) {
+                        invalidClaim = {
+                            validatorId: validator.id,
+                            reason: validationRes.reason,
+                        };
```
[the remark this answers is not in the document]
rishabhpoddar: @porcellus missed this?

examples/with-claims/src/test.context.jsx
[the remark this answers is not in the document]
porcellus: We've touched on it on a call, but basically, this file is there to support the "log view" for the example. This is mainly intended to demo/visualize the events/API calls/etc. I made this to show how claims work under the hood. 
I think this could be useful in general, but not really as an example to follow, so I'll remove this later.

lib/ts/recipe/session/recipe.ts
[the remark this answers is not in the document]
porcellus: `redirectToUrl` should handle it the same way this works for auth recipes.

lib/ts/recipe/session/types.ts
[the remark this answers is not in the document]
porcellus: Well, the events we get from the session recipe can't contain the claim checking info, the required claims may be different locally. 
There are no implications for the user, they get the same events as before. Essentially the users will get this:
```ts
type RecipeEventWithSessionContext = {
  // ....
  sessionContext: {
    doesSessionExist: boolean;
    userId: string;
    accessTokenPayload: any;
  }
}
```

lib/ts/recipe/session/sessionAuth.tsx
porcellus: It's not really good practice to make calls (and show buttons) that you know will fail. 
We may not have enough information (so the error only appears on the server-side), which should be rare or the information we have is outdated so we should re-validate the claims.

lib/ts/recipe/session/sessionAuth.tsx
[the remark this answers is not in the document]
porcellus: No, but we've discussed this on a call before, and I thought we'd address it in a separate PR with the email verification claims. So basically there are a few options:
1. Special casing this if it's added by the other recipe. This can be done by adding a version of the claim as a global claim that can tell the session verification recipe to redirect. The `SuperTokensWrapper` will have to disable this to avoid a redirection loop.
3. Adding specific config into the session auth wrapper to do these kinds of redirections (or do them if `requireAuth` is set to true)

lib/ts/recipe/session/index.ts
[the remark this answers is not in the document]
porcellus: We've discussed removing recipe-specific auth components and simply redirecting users to websiteBasePath instead of using the recipe id of the current auth component.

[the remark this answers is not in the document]
rishabhpoddar: We do not need to expose this to the user from the index.ts file

lib/ts/recipe/session/sessionAuth.tsx
```diff
@@ -98,79 +111,75 @@ const SessionAuth: React.FC<PropsWithChildren<Props>> = ({ children, ...props })
...
                 case "SESSION_CREATED":
-                    setContext(event.sessionContext);
-                    return;
                 case "REFRESH_SESSION":
-                    setContext(event.sessionContext);
-                    return;
                 case "ACCESS_TOKEN_PAYLOAD_UPDATED":
-                    setContext(event.sessionContext);
-                    return;
                 case "SIGN_OUT":
```
rishabhpoddar, line 138: do we need to do claim validation if this is the true as well?

rishabhpoddar: (Maybe make this similar to UNAUTHORISED case, but without calling the callbacks

lib/ts/recipe/session/sessionAuth.tsx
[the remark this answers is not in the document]
rishabhpoddar: Keeping this open for email veirifcation PR

rishabhpoddar (approved):

examples/with-claims/api/claims/EmailVerifiedClaim.ts
```diff
@@ -0,0 +1,30 @@
+import { SessionClaimValidator } from "supertokens-node/recipe/session";
+import { BooleanClaim } from "supertokens-node/recipe/session/claims";
+import { isEmailVerified } from "supertokens-node/recipe/emailpassword";
+
+class EmailVerifiedClaimClass extends BooleanClaim {
```
rishabhpoddar, line 5: We need to make this example app much simpler to consume and understand:
- We will remove this claim since it will be prebuilt
- Rename the claims folder to `customClaims`
- RolesClaim should also be removed since it should be inbuilt.
- The custom APIs that use claims should be fewer in number - for example, we do not need to show refreshing or email verification claim API.

Keeping this comment open until everything else is done, and we can work on cleaning up the example app as the last step.

rishabhpoddar (approved):

lib/ts/recipe/session/sessionAuth.tsx
```diff
@@ -57,19 +52,29 @@ const SessionAuth: React.FC<PropsWithChildren<Props>> = ({ children, ...props })
...
+    const session = useRef<Session>();
     const userContext = useUserContext();
 
+    const redirectToLogin = useCallback(() => {
+        if (props.redirectToLogin !== undefined) {
+            props.redirectToLogin();
+        } else {
+            void session.current!.redirectToAuthWithRedirectToPath(
+                SuperTokens.getInstanceOrThrow().getReactRouterDomWithCustomHistory()?.useHistoryCustom()
+            );
```
rishabhpoddar, line 68: We should change this to call a top level getRedirectionURL function (on supertokens.init level) with a context like:
```
{
   action: "TO_AUTH",
   showSignIn: boolean | undefined
}
```

And exposed `SuperTokens.redirectToAuth({...})` takes the same stuff as what we have right now for individual recipes. It should also set the right query params from the return of `getRedirectionURL` (if the user has not set them already)

We should remove the recipe level redirectToAuth functions, and the SIGN_IN_UP action associated with them.

lib/ts/recipe/session/sessionAuth.tsx
```diff
@@ -23,25 +23,20 @@ import { RecipeEventWithSessionContext, SessionContextType } from "./types";
...
-
-type PropsWithAuth = {
-    requireAuth: true;
-    redirectToLogin: () => void;
-};
-
-type Props = (PropsWithoutAuth | PropsWithAuth) & {
+export type SessionAuthProps = {
+    requireAuth?: boolean;
+    redirectToLogin?: () => void;
```
rishabhpoddar, line 31: We should remove this prop.

rishabhpoddar (changes requested):

lib/ts/recipe/session/sessionContext.ts
```diff
@@ -6,6 +6,7 @@ const SessionContext = React.createContext<SessionContextType & { isDefault: boo
     isDefault: true,
 });
 
+// ts-prune-ignore-next
 export function isDefaultContext(sessionContext: SessionContextType & { isDefault: boolean }): boolean {
```
rishabhpoddar, line 10: need to add `sessionContext.getClaimValue` function as well.

rishabhpoddar (changes requested):

examples/with-phone-password/README.md
```diff
@@ -46,27 +46,31 @@ This demo app uses the EmailPassword and Passwordless recipes to achieve the aut
 
 ### Frontend
 
+-   Add a custom `PhoneVerifiedClaim`.
+-   Override `getGlobalClaimValidators` in the Session recipe to add a validator for `PhoneVerifiedClaim`. This will add an error into `invalidClaims` in the session context of components wrapped by `SessionAuth` if PhoneVerifiedClaim` is missing or set to false.
```
rishabhpoddar, line 50:
```suggestion
-   Override `getGlobalClaimValidators` in the Session recipe to add a validator for `PhoneVerifiedClaim`. This will add an error into `invalidClaims` in the session context of components wrapped by `SessionAuth` if `PhoneVerifiedClaim` is missing or set to false.
```

examples/with-phone-password/README.md
```diff
@@ -46,27 +46,31 @@ This demo app uses the EmailPassword and Passwordless recipes to achieve the aut
...
--   Provide `getRedirectionURL` to the EmailPassword.init function to redirect the user to the second login challenge in case the first challenge is complete.
--   We disable the default UI of passwordless recipe so that we can render what we want on the /verify-phone route
--   We override the enter phone number component for passwordless recipe to send the OTP to the phone number directly - as opposed to it asking the user to enter it again.
--   We override the "Change phone number button" in the enter OTP screen with a custom button which will logout the user and take them back to the first login challenge.
+-   Added text translation to password reset forms and sign-in / up forms
+-   We disable the default UI of passwordless recipe and render a modified version on the `/verify-phone` route
+-   We create a route for the second challenge (to verify phone number via OTP) on `/auth/verify-phone`. On this route, we send the OTP and ask the user to enter the OTP.
```
rishabhpoddar, line 55:
```suggestion
-   We disable the default UI of passwordless recipe and render a modified version on the `/auth/verify-phone` route. On this route, we send the OTP and ask the user to enter the OTP.
```

examples/with-phone-password/README.md
```diff
@@ -46,27 +46,31 @@ This demo app uses the EmailPassword and Passwordless recipes to achieve the aut
...
+-   Provide `getRedirectionURL` to the `EmailPassword.init` function to redirect the user to the second login challenge in case the first challenge is complete.
+-   We override the enter phone number component for the passwordless recipe to send the OTP to the phone number directly - as opposed to it asking the user to enter it again.
```
rishabhpoddar, line 57:
```suggestion
-   We override the enter phone number component for the passwordless recipe to send the OTP to the phone number directly - as opposed to it asking the user to enter it again (during sign in).
```

rishabhpoddar (changes requested):

rishabhpoddar (approved):

rishabhpoddar (approved):

PR #402, opened 2022-03-22
feat!: Integrate passwordless recipe from web-js SDK

nkshah2:
## Summary of change

- Integrate passwordless recipe from web-js SDK
- Modify build-pretty script to run pretty twice

## Related issues

## Test Plan

Screenshot 2022-03-22 at 5 21 03 PM

## Documentation changes

WIP

    -  ~~ In `package.json`~~
    -  ~~ In `lib/ts/version.ts`~~
    -   ~~To know which one it is, run find the latest released tag (`git tag`) in the format `vX.Y.Z`, and then find the latest branch (`git branch --all`) whose `X.Y` is greater than the latest released tag.~~
    -   ~~If no such branch exists, then create one from the latest released branch.~~

-   [x] Update CHANGELOG
-   [x] Verify size limit


nkshah2:
> There are a few changes to web-js PR, so moving this back to in progress as well

Updated the web-JS PR and updated this one, all tests still pass

rishabhpoddar:
I think it would be better if @porcellus reviews this instead of me - since he wrote this recipe in the first place. After his approval, I can have a brief look.

lib/ts/recipe/passwordless/recipeImplementation.ts
```diff
@@ -1,295 +1,104 @@
...
-                    version: 1,
-                    ...input,
-                })
-            );
+        clearLoginAttemptInfo: function (input) {
+            return webJsImplementation.clearLoginAttemptInfo.bind(this)(input);
         },
-        clearLoginAttemptInfo: function (): void {
-            removeFromLocalStorage(PASSWORDLESS_LOGIN_ATTEMPT_INFO_STORAGE_KEY);
+        didLoginAttemptInfoChangeAfterResend: function (input) {
```
porcellus, line 100: - This seems to be unused
- I think you could drop "AfterResend" from the name.

package.json
```diff
@@ -85,7 +85,7 @@
         "build": "cd lib && rm -rf build && npx tsc -p tsconfig.json",
         "watch": "cd lib && npx tsc -w -p tsconfig_dev.json",
         "pretty": "npx pretty-quick .",
-        "build-pretty": "npm run build && npm run pretty",
+        "build-pretty": "npm run build && npm run pretty && npm run pretty",
```
porcellus, line 88: why are we running pretty twice?

lib/ts/recipe/passwordless/utils.ts
```diff
@@ -158,3 +159,28 @@ export function defaultGuessInternationPhoneNumberFromInputPhoneNumber(
...
+          flowType: PasswordlessFlowType;
+          contactInfo: string;
+          contactMethod: "EMAIL" | "PHONE";
+          lastResend: number;
+          redirectToPath?: string;
+      }
+    | undefined
+> {
+    return await input.recipeImplementation.getLoginAttemptInfo<{
+        contactInfo: string;
```
porcellus, line 179: I'd prefer if this was extracted into `types.ts` and reused everywhere we with `getLoginAttemptInfo`.

lib/ts/recipe/passwordless/utils.ts
```diff
@@ -158,3 +159,28 @@ export function defaultGuessInternationPhoneNumberFromInputPhoneNumber(
     }
     return value;
 }
+
+export async function getLoginAttemptInfoFromStorage(input: {
```
porcellus, line 163: Is this util here just to help with types?

porcellus (changes requested):

lib/ts/recipe/passwordless/components/features/linkClickedScreen/index.tsx
```diff
@@ -31,6 +31,7 @@ import { ComponentOverrideContext } from "../../../../../components/componentOve
 import { defaultTranslationsPasswordless } from "../../themes/translations";
 import { useEffect } from "react";
 import { useUserContext } from "../../../../../usercontext";
+import STGeneralError from "supertokens-web-js/lib/build/error";
```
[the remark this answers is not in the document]
nkshah2: All `supertokens-web-js/lib/build` imports will be refactored in a future PR

lib/ts/recipe/passwordless/recipeImplementation.ts
```diff
@@ -1,295 +1,104 @@
...
-                //     return input.config.preAPIHook({
-                //         ...context,
-                //         action: "PHONE_NUMBER_EXISTS",
-                //         userContext: input.userContext,
-                //     });
-                // }
-            );
-
-            return response.exists;
+        getPreAuthSessionIdFromURL: function (input) {
```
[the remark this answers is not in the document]
nkshah2: I agree that they can be clubbed but for the sake of clarity I prefer to have them as separate functions (im not super biased towards this so if you feel strongly about having them together I can make the change)

They are used in the web-js SDK, auth-react uses the same recipe interface as web-js now and we want to allow the users to be able to override them if they want so they are simply declared in auth-react

lib/ts/recipe/passwordless/recipeImplementation.ts
nkshah2: `didLoginAttemptInfoChangeAfterResend` will get removed (I just havent done it because I wanted to club review changes in one push)

package.json
nkshah2: There was this annoying issue where after running pretty once pretty-check would still fail for some files (the same issue happened in supertokens-node as well). There was no clear reason as to why it was happening and running pretty again solved it so went with this approach for now.

lib/ts/recipe/passwordless/utils.ts
nkshah2: The util function `getLoginAttemptInfoFromStorage` was introduced to solve this problem (I must have missed refactoring in this case). Ive changed the util function to use types declared in `types.ts` along with changing calls to `setLoginAttemptInfo` to use the types as well

lib/ts/recipe/passwordless/utils.ts
nkshah2: Yes and to prevent the need to pass a generic every time we call `getLoginAtemptInfo` on the recipe interface (also refer to the comment before this oen)

lib/ts/recipe/passwordless/recipeImplementation.ts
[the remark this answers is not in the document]
porcellus: No strong preference, it's fine the way it is.

package.json
porcellus: Hmm, what changes between the two runs? We should investigate this a bit. Running it twice is fine for the time being, but we should create an issue/TODO about this.

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
```diff
@@ -264,81 +267,122 @@ export const SignInUpFeature: React.FC<
...
 ): RecipeInterface {
     return {
-        ...originalImpl,
+        ...recipe.recipeImpl,
         createCode: async (input) => {
             let contactInfo;
             if ("email" in input) {
                 contactInfo = input.email;
+
+                const validationRes = await recipe.config.validateEmailAddress(input.email);
```
porcellus, line 281: This is also done in the forms where we are using createCode.

lib/ts/recipe/passwordless/components/themes/signInUp/linkSent.tsx
```diff
@@ -67,9 +66,6 @@ const PasswordlessLinkSent: React.FC<LinkSentThemeProps> = (props) => {
                 }, 2000);
             } else {
                 setStatus("ERROR");
-                if (response.status === "GENERAL_ERROR") {
-                    props.onError(response.message);
```
porcellus, line 71: Shouldn't we make this call in the catch block?

porcellus (changes requested):

lib/ts/recipe/passwordless/components/features/linkClickedScreen/index.tsx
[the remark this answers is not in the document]
nkshah2: Yes

package.json
nkshah2: So one case is that the first time we run pretty, the thirdpartyemailpassword/index.js changes in a way that pretty-check will fail. To fix this you would have to split the code in multiple lines but when you do that pretty changes it back to one line which makes it too long for pretty-check.

This looks like a case where prettier rules conflict with each other and one takes precedence. This will eventually be a problem in all TS SDKs where we use prettier.

lib/ts/recipe/passwordless/utils.ts
nkshah2: Well the idea was to maintain the recipe function name in the util function name to make it clear that they are co-related (`getLoginAttemptInfo` being the recipe function). There is an argument that I could have used the same name and done it that way but I think that might be confusing

lib/ts/recipe/passwordless/components/themes/signInUp/linkSent.tsx
nkshah2: it is being made

```ts
try {
    ...

    if (response.status === "OK") {
        setStatus("LINK_RESENT");
        ...
    } else {
        setStatus("ERROR");
    }
} catch (e) {
    setStatus("ERROR");
}
```

nkshah2: So the change here was just to remove the GENERAL_ERROR handling, setError was being called originally as well so it probably isnt showing up in git diff

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
nkshah2: This validation has also been added to (or modified to throw STGeneralError in):
- `themes/signInUp/emailForm.tsx` (email only)
- `themes/signInUp/emailOrPhoneForm.tsx`
- `themes/signInUp/phoneForm.tsx` (phone only)

porcellus: That's my point: this is done both before and during `createCode`. If it's done during the call it should be removed from those forms I think.

lib/ts/recipe/passwordless/utils.ts
porcellus: I think the same name might be fine, or you could add `ForReact` as a suffix because I think FromStorage is not 100% correct: it indicates local (or session) storage but it could just as well be customized. 
Anyway, this is fine as it is if you prefer it this way.

lib/ts/recipe/passwordless/components/themes/signInUp/linkSent.tsx
porcellus: I meant the `props.onError(response.message);`

porcellus (approved):
LGTM

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
rishabhpoddar: @porcellus @nkshah2 this comment is not resolved.

lib/ts/recipe/passwordless/index.ts
```diff
@@ -17,8 +17,9 @@ import { UserInput } from "./types";
 
 import Passwordless from "./recipe";
 import PasswordlessAuth from "./passwordlessAuth";
-import { GetRedirectionURLContext, PreAPIHookContext, OnHandleEventContext, RecipeInterface } from "./types";
+import { GetRedirectionURLContext, PreAPIHookContext, OnHandleEventContext } from "./types";
 import SignInUpThemeWrapper from "./components/themes/signInUp";
+import { RecipeInterface } from "supertokens-web-js/recipe/passwordless";
 
 export default class Wrapper {
     static init(config: UserInput) {
```
rishabhpoddar, line 25: You need to expose all the functions of this recipe?

Please make sure that all the recipes you have implemented so far have done this.

lib/ts/recipe/passwordless/components/themes/signInUp/linkSent.tsx
```diff
@@ -67,12 +67,13 @@ const PasswordlessLinkSent: React.FC<LinkSentThemeProps> = (props) => {
...
-                }
             }
         } catch (e) {
             setStatus("ERROR");
+
+            if (STGeneralError.isThisError(e)) {
+                props.onError(e.message);
+            }
```
rishabhpoddar, line 76: How does this work? We first do `setStatus("ERROR")` which I assume shows something went wrong, and then we show the user specific error message if `STGeneralError.isThisError(e)`?

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
```diff
@@ -264,81 +267,112 @@ export const SignInUpFeature: React.FC<
...
+                if (STGeneralError.isThisError(e)) {
+                    dispatch({ type: "setError", error: e.message });
                 }
-            } else if (res.status === "RESTART_FLOW_ERROR") {
-                await originalImpl.clearLoginAttemptInfo();
 
-                dispatch({ type: "restartFlow", error: "ERROR_SIGN_IN_UP_RESEND_RESTART_FLOW" });
-            } else if (res.status === "GENERAL_ERROR") {
-                dispatch({ type: "setError", error: res.message });
+                throw e;
```
rishabhpoddar, line 344: Do you want to throw an error even if `STGeneralError.isThisError(e)` is `true`?

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
```diff
@@ -264,81 +267,112 @@ export const SignInUpFeature: React.FC<
 export default SignInUpFeature;
 
 function getModifiedRecipeImplementation(
-    originalImpl: RecipeInterface,
+    recipe: Recipe,
```
rishabhpoddar, line 270: You should pass the recipe implementation here. And call it `originalImpl`. Just to minimise the number of changes.

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
```diff
@@ -264,81 +267,112 @@ export const SignInUpFeature: React.FC<
...
+                        userContext: input.userContext,
+                    });
+
+                    dispatch({ type: "restartFlow", error: "ERROR_SIGN_IN_UP_RESEND_RESTART_FLOW" });
+                }
+                return res;
+            } catch (e) {
+                if (STGeneralError.isThisError(e)) {
+                    dispatch({ type: "setError", error: e.message });
                 }
```
rishabhpoddar, line 342: How come this is not being done in the other functions in this file (like `consumeCode`)

lib/ts/recipe/emailpassword/components/library/formBase.tsx
```diff
@@ -151,8 +156,17 @@ export const FormBase: React.FC<FormBaseProps<any>> = (props) => {
...
+                     * throwing STGeneralError, we need to handle this here too
+                     */
+                    if (unmounting.current.signal.aborted) {
+                        return;
+                    }
+
+                    handleFieldStateChange(fieldUpdates);
+
+                    props.onError(e.message);
                     return;
```
rishabhpoddar, line 170: In general, doing a return from checking `STGeneralError.isThisError(e)` is not good. What if `STGeneralError.isThisError(e)` is not true, and we just want to do  `props.onError("SOMETHING_WENT_WRONG_ERROR");` and continue to do something else?

So I would change this return everywhere to be like:
```
if (STGeneralError.isThisError(e)) {
   // handle general error
} else {
  // whatever the old code was doing in case of catching an exception
}
```

rishabhpoddar (changes requested):

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
nkshah2: @rishabhpoddar This has been solved but just to confirm, @porcellus this is good to be resolved right?

lib/ts/recipe/passwordless/components/themes/signInUp/linkSent.tsx
nkshah2: Thats correct, setState("ERROR") indicates that there is an error and in the case of an STGeneralError we want the UI to show the appropriate error message

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
nkshah2: Yes, this is the modified recipe implementation. When this function is called the errors get handled there.

lib/ts/recipe/passwordless/index.ts
nkshah2: Done, also verified in all other recipes

lib/ts/recipe/passwordless/index.ts
```diff
@@ -50,6 +52,167 @@ export default class Wrapper {
...
+                linkCode,
+                preAuthSessionId,
+            };
+        }
+
+        return recipe.recipeImpl.consumeCode({
+            userContext: normalisedUserContext,
+            options: input.options,
+            ...additionalParams,
+        });
```
rishabhpoddar, line 187: Whilst this is fine, it's being repeated in web-js SDK as well. Is there anyway we can not have this duplicate code across the two SDKs?

I know that we can't just call the the web-js' exposed function here cause maybe the user has overrides the recipe functions of this SDK. But a solution could be that in webjs, we have a file that has all these functions but they all take a recipe interface as an input too. And then those functions use that recipe interface for the logic instead of `Passwordless.getInstanceOrThrow(). recipeImpl `.

Then this SDK can call those functions with `Passwordless.getInstanceOrThrow(). recipeImpl` as an input and the web-js' index file functions can call those functions with its `Passwordless.getInstanceOrThrow(). recipeImpl` as input

lib/ts/recipe/passwordless/components/features/linkClickedScreen/index.tsx
```diff
@@ -52,15 +53,54 @@ const LinkClickedScreen: React.FC<PropType> = (props) => {
...
+                          status: "INCORRECT_USER_INPUT_CODE_ERROR" | "EXPIRED_USER_INPUT_CODE_ERROR";
+                          failedCodeInputAttemptCount: number;
+                          maximumCodeInputAttempts: number;
+                          fetchResponse: Response;
+                      }
+                    | {
+                          status: "RESTART_FLOW_ERROR";
+                          fetchResponse: Response;
+                      }
+                    | undefined;
```
rishabhpoddar, line 73: Do you need to define the full types here?

lib/ts/recipe/thirdparty/components/features/signInAndUpCallback/index.tsx
```diff
@@ -41,9 +42,42 @@ class SignInAndUpCallback extends PureComponent<PropType, unknown> {
...
+                      status: "OK";
+                      user: UserType;
+                      createdNewUser: boolean;
+                      fetchResponse: Response;
+                  }
+                | {
+                      status: "NO_EMAIL_GIVEN_BY_PROVIDER";
+                      fetchResponse: Response;
+                  }
+                | undefined;
```
rishabhpoddar, line 56: Do you need to define full type here?

lib/ts/recipe/thirdparty/components/themes/signInAndUp/providersForm.tsx
```diff
@@ -29,23 +29,39 @@ export const ThirdPartySignInAndUpProvidersForm: React.FC<SignInAndUpThemeProps>
...
+                    generalError = e;
+                } else {
+                    throw e;
+                }
             }
-        } catch (err) {
-            if (STGeneralError.isThisError(err)) {
+
+            if (generalError !== undefined) {
                 return props.dispatch({
```
rishabhpoddar, line 51: Well.. you are still having to `return` from these types of general errors from the function which kind of beats the point of this refactor (whose original purpose was not to return if the older general error handling didn't return). If we must always return, then I prefer it being in the try catch like how it was done before this PR..

So overall, the point of this is that:
- If previously we were handling general error by checking the status, that code flow should be unchanged and only the way we detected if it was a general error would change.
- If previously we didn't have a general error status for an API call, we should now handle it in the try / catch block.

If you want to maintain that we handle general error in try catch everywhere, you need to make sure that if previously we handled general error without a return, the try / catch doesn't have a return then either.

rishabhpoddar (changes requested):

lib/ts/recipe/passwordless/components/features/linkClickedScreen/index.tsx
nkshah2: I was having this weird issue where typescript would complain about `createdUser` not being present even if status was OK. Adding the types resolved this issue.

lib/ts/recipe/thirdparty/components/features/signInAndUpCallback/index.tsx
nkshah2: Same as the comment above

lib/ts/recipe/thirdparty/components/themes/signInAndUp/providersForm.tsx
nkshah2: Looking at it again, this did not need to return early. Refactored it to handle it the same way it is done in other places

lib/ts/recipe/passwordless/index.ts
nkshah2: I dont like the idea of passing `recipeImpl`, but I also cant think of any other way of avoiding duplication without adding some complexity. In this case I think it is fine to leave it as is

rishabhpoddar: ok

rishabhpoddar (approved):

PR #422, opened 2022-04-13
refactor: Use web-js input type for session recipe

nkshah2:
## Summary of change
- Uses input types for session recipe from supertokens-web-js instead of redeclaring them

## Related issues
-   https://github.com/supertokens/for-zenhub/issues/28

## Related PRs (These should be merged before this one)
- https://github.com/supertokens/supertokens-web-js/pull/24

## Test Plan

Tests from Github actions should pass

## Documentation changes

None

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

-   [ ] Verify Size Limit

### DIFF
--- a/lib/ts/recipe/session/types.ts
+++ b/lib/ts/recipe/session/types.ts
@@ -13,8 +13,7 @@
  * under the License.
  */
 
-import { RecipeInterface } from "supertokens-website";
-import OverrideableBuilder from "supertokens-js-override";
+import { UserInput as WebJSInputType } from "supertokens-web-js/lib/build/recipe/session/types";
 
 export type RecipeEvent =
     | {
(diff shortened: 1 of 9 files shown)


lib/ts/recipe/session/types.ts
```diff
@@ -29,26 +28,8 @@ export type RecipeEvent =
...
-    }) => Promise<{ url: string; requestInit: RequestInit }>;
-    onHandleEvent?: (event: RecipeEvent) => void;
-    override?: {
-        functions?: (
-            originalImplementation: RecipeInterface,
-            builder?: OverrideableBuilder<RecipeInterface>
-        ) => RecipeInterface;
-    };
+export type InputType = WebJSInputType & {
+    onHandleEvent?: (event: RecipeEventWithSessionContext) => void;
```
rishabhpoddar, line 32: Where is the postAPIHook? Am i confusing with something else?

rishabhpoddar (changes requested):

lib/ts/recipe/session/types.ts
nkshah2: `WebJSInputType` contains both pre and post API hooks. The only different in webjs types and this one is the onHandleEvent which needs to session context so ive commented it

rishabhpoddar (approved):

PR #424, opened 2022-04-14
refactor: Remove Querier from supertokens-auth-react

nkshah2:
## Summary of change
- Removes querier from auth-react and refactors imports to use the one from web-js
- Removes `supported_fdi` from the versions.ts file

## Related issues

## Test Plan

Existing test cases should pass (From the PR action)

## Documentation changes

None

-   [ ] ~~Changelog has been updated~~
    -   ~~In `package.json`
    -   ~~In `lib/ts/version.ts`~~

-   Verify size limit


test/unit/querier.test.ts
```diff
@@ -10,8 +10,6 @@ describe("Querier", () => {
             apiBasePath: new NormalisedURLPath("/"),
             apiDomain: new NormalisedURLDomain("http://api.example.com"),
             appName: "test",
-            websiteBasePath: new NormalisedURLPath("/"),
-            websiteDomain: new NormalisedURLDomain("http://example.com"),
         });
```
rishabhpoddar, line 13: Shouldn't this whole file be moved to web-js as well?

rishabhpoddar (changes requested):
Also you need to change the PR template. It mentions that you need to modify the array in the version.ts file if changing the FDI. This is no longer true..

test/unit/querier.test.ts
nkshah2: Done, refer to: https://github.com/supertokens/supertokens-web-js/pull/25

rishabhpoddar (approved):

Issue #433, opened 2022-04-27
Fix example apps


rishabhpoddar:
- [x] Update to latest version of the SDK in the example apps
- [x] Update to react 18
- [x] Fix account linking example (store mapping info like appUserId -> {recipeId1: stUserId, recipeId2: stUserId, ...}
- [x] Get all example apps to work
- [x] Remove "with-next-one-login-many-subdomains" example app
- [x] Remove blitz example app
- [x] Update https://github.com/supertokens/supertokens-auth-react/tree/master/examples/with-thirdpartypasswordless-electron to use react 18

Issue #428, opened 2022-04-18
Add compatibility with React 18


rishabhpoddar:
- [x] In react strict mode and in dev mode, `useEffect` / `componenDidMount` is called twice as well: https://reactjs.org/blog/2022/03/29/react-v18.html#new-strict-mode-behaviors. So some of the feature components we have (like email verification or sign in up callback) which do an API call on component mount, will need to change so that the API call happens only once.

PR #438, opened 2022-05-06
docs: fix example apps

porcellus:
## Summary of change

Fixes minor issues in the example apps:
- Removed unnecessary logs
- Expanded READMEs
- Updated `with-next-iframe` to be consistent with normal next example
- `with-one-login-many-subdomains` & `with-one-login-per-subdomain`:
  - Fixed domain whitelist regex
  - Explicitly adding JSON body-parser.
- `with-thirdpartyemailpassword-passwordless`: updated how we store users and storing recipe specific user ids

## Related issues

-  #433 
- #428

## Test Plan

Only examples are updated

## Documentation changes

Only examples are updated

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`


examples/with-hasura-thirdpartyemailpassword/README.md
```diff
@@ -29,7 +29,7 @@ npm install
 ngrok http 3001
```
rishabhpoddar, line 29: Shouldn't the example apps be updated to use react 18? You can change the supertokens-auth-react dependency in all the example apps to point to the branch which has the react 18 compatibility. Then once we release it, we can change it to use npm.

rishabhpoddar (changes requested):

examples/with-hasura-thirdpartyemailpassword/README.md
porcellus: They have been updated in #439

PR #440, opened 2022-05-06
test: update tests to react 18

porcellus:
## Summary of change

Update tests to use react 18

## Related issues

-   #433 
- #428

## Test Plan

Test changes

## Documentation changes

Test changes only

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`


test/end-to-end/passwordless.test.js
```diff
@@ -1408,6 +1427,7 @@ export function getPasswordlessTestCases({ authRecipe, logId }) {
                     "ST_LOGS SESSION OVERRIDE ADD_FETCH_INTERCEPTORS_AND_RETURN_MODIFIED_FETCH",
                     "ST_LOGS SESSION OVERRIDE ADD_AXIOS_INTERCEPTORS",
                     `ST_LOGS ${logId} OVERRIDE GET_LOGIN_ATTEMPT_INFO`,
+                    `ST_LOGS ${logId} OVERRIDE GET_LOGIN_ATTEMPT_INFO`,
```
rishabhpoddar, line 1430: Why are these events getting called twice?

examples/for-tests/package.json
```diff
@@ -4,8 +4,8 @@
     "private": true,
     "dependencies": {
         "axios": "^0.21.0",
```
rishabhpoddar, line 6: Are we testing with react 17 and with react 18?

rishabhpoddar (changes requested):

examples/for-tests/package.json
porcellus: Not yet, I'll add that in a separate PR. (Most likely next week)

test/end-to-end/passwordless.test.js
porcellus: They are part of the state initialization of the passwordless signinup feature component. They are called twice because the state wasn't updated before the remount (unmounted before the `load` dispatch). 
We could add a separate bit of state with useRef that could avoid this test change, but it doesn't feel appropriate:
- These things are OK to be called multiple times. (For API calls, error and success handlers can be called multiple times as well since they are concerned with the local UI state)
- We'd be adding an extra state to fix behavior that is only an issue during testing.

examples/for-tests/package.json
rishabhpoddar: okay.

test/end-to-end/passwordless.test.js
rishabhpoddar: I think you should put it as part of this PR, or on top of this PR.

test/end-to-end/signin-rrdv5.test.js
```diff
@@ -61,7 +61,7 @@ import { EMAIL_EXISTS_API, SIGN_IN_API, TEST_CLIENT_BASE_URL, TEST_SERVER_BASE_U
 /*
  * Tests.
  */
-describe("SuperTokens SignIn with react router dom v5", function () {
+describe.skip("SuperTokens SignIn with react router dom v5", function () {
```
rishabhpoddar, line 64: Need to remove `skip` and run this is using react 17

rishabhpoddar (changes requested):

test/end-to-end/passwordless.test.js
rishabhpoddar: > They are called twice because the state wasn't updated before the remount (unmounted before the load dispatch).

Could you explain this more? Also, will the API calls for consuming the third party code be called twice and the first one is getting aborted?

porcellus: Yep:
API calls are behind `onMountAPICall` which only calls the API only once and saves the resulting promise. This then either calls a response or an error handler which can be called multiple times, because:
- These may be interrupted (they are after a potentially long async operation) while making changes to local state.
- They should be only concerned with UI state, so calling them twice shouldn't be a problem (except for the tests)
That's the cause of redirect handlers being called twice.

`GET_LOGIN_ATTEMPT_INFO` is called twice because it's part of initial state loading and I didn't consider it to be an API call (it just reads localstorage), so I didn't put it behind `onMountAPICall`. It's should not cause a problem and it actually sets the state immediately after `getLoginAttemptInfo` returns.

examples/for-tests-react-16/package.json
```diff
@@ -0,0 +1,37 @@
...
+        "axios": "^0.21.0",
+        "react": "^16.14.0",
+        "react-dom": "^16.14.0",
+        "react-router-dom": "^6.2.1",
+        "react-router-domv5": "npm:react-router-dom@^5.2.0",
+        "react-scripts": "3.4.3"
+    },
+    "scripts": {
+        "link": "cd ../../ && SRC=./examples/for-tests-react-16 npm run prep-test-app",
+        "start": "PORT=3031 react-scripts start",
```
rishabhpoddar, line 15: Consider running npm run link here as well (before)

rishabhpoddar (changes requested):

rishabhpoddar (approved):

PR #458, opened 2022-05-20
fix(examples, with-localstorage): update headers on the response obje…

p6l-richard:
## Summary of change

The `getSession` overwrite errors out as the input parameter for updating the headers is wrong.

This PR fixes this by passing the response object as the parameter instead of the session.

## Related issues

image

## Test Plan

This seems to be untested


rishabhpoddar:
Thanks @p6l-richard

PR #462, opened 2022-05-24
docs: adds Angular example app

jscyo:
## Summary of change

Adds angular example app

## Related issues
- https://github.com/supertokens/for-zenhub/issues/123

## Test Plan

## Documentation changes

-   [ ] ~`frontendDriverInterfaceSupported.json` file has been updated (if needed)~
    -   ~Along with the associated array in `lib/ts/version.ts`~
-   [ ] ~Changes to the version if needed~
    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`
-   [ ] ~Issue this PR against the latest non released version branch.~
-   [ ] ~If added a new recipe interface, then make sure that the implementation of it uses NON arrow functions only (like `someFunc: function () {..}`).~
-   [ ]~ If I added a new recipe, I also added the recipe entry point into the `size-limit` section of `package.json` with the size limit set to the current size rounded up.~


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 -   Adds a SuperTokens + Supabase example app
 -   Adds an example app with svelte
 -   Adds phone number and password demo app
+-   Adds an example app with Angular
```
rishabhpoddar, line 13:
```suggestion
-   Adds an example app with Angular + React
```

examples/with-angular-thirdpartyemailpassword/README.md
```diff
@@ -0,0 +1,43 @@
+![SuperTokens banner](https://raw.githubusercontent.com/supertokens/supertokens-logo/master/images/Artboard%20%E2%80%93%2027%402x.png)
+
+# Angular Example app using SuperTokens React library
+
+# SuperTokens ThirdpartyEmailPassword demo app with Angular
```
rishabhpoddar, line 5: Two titles?

rishabhpoddar: Also mention Angular with React

examples/with-angular-thirdpartyemailpassword/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+This demo app demonstrates the following use cases:
+
+-   Login / sign up with SuperTokens
+-   Logout
+-   Forgot password flow
+-   Session management
+
+This also showcases how you can use the supertokens-auth-react SDK inside a svelte app.
```
rishabhpoddar, line 14: svelte?

examples/with-angular-thirdpartyemailpassword/README.md
```diff
@@ -0,0 +1,43 @@
...
+
```
rishabhpoddar, line 15: Add more details about how it actually works.. that on /auth/* route we use react, but in others, we use angluar

examples/with-angular-thirdpartyemailpassword/server.js
```diff
@@ -0,0 +1,66 @@
+const express = require("express");
```
rishabhpoddar, line 1: Convert this project to use TS

rishabhpoddar: For the server, you can run it using ts-node (instead of node), so you should add ts-node as a packaege.json dependency as well.

ts-node allows you to directly run ts files for the server

rishabhpoddar (changes requested):

examples/with-angular-thirdpartyemailpassword/src/app/home/home.component.ts
```diff
@@ -0,0 +1,37 @@
+import { AfterViewInit, Component, OnChanges, OnDestroy, SimpleChanges } from "@angular/core";
+
+import * as Session from "supertokens-auth-react/recipe/session";
+import { signOut } from "supertokens-auth-react/recipe/thirdpartyemailpassword";
```
rishabhpoddar, line 4: This is an application route.. ideally it should not use supertokens-auth-react and only use supertokens-website.

examples/with-angular-thirdpartyemailpassword/src/app/auth/auth.component.ts
```diff
@@ -0,0 +1,24 @@
...
+    title = "angularreactapp";
+
+    public rootId = "rootId";
+
+    ngAfterViewInit() {
+        this.render();
+    }
+
+    private render() {
+        ReactDOM.render(React.createElement(SuperTokensReactComponent), document.getElementById(this.rootId));
```
rishabhpoddar, line 22: How do you know that this is the right way to do this?

rishabhpoddar (changes requested):

examples/with-angular-thirdpartyemailpassword/src/app/auth/supertokens.tsx
```diff
@@ -0,0 +1,37 @@
...
+            signInAndUpFeature: {
+                providers: [Github.init(), Google.init()],
+            },
+            emailVerificationFeature: {
+                mode: "REQUIRED",
+            },
+        }),
+        Session.init(),
+    ],
+});
```
rishabhpoddar, line 26: You are calling this only in the auth route. What if this route is never loaded? Therefore this will never be called, and so the home route will break.

If this is being called even if the user doesn't navigate to the auth route, it means that you are loading supertokens-auth-react on each page load anyway - which breaks the point of the code splitting.. how is it working?

rishabhpoddar (changes requested):

examples/with-angular-thirdpartyemailpassword/src/app/auth/auth.component.ts
rishabhpoddar: @jscyo you missed this.

jscyo: Looked at a number of implementations of how people were using React components in angular and used the same method. I've added comments explaining what each lifecycle hook does

examples/with-angular-thirdpartyemailpassword/src/app/app.component.ts
```diff
@@ -0,0 +1,16 @@
+import { Component } from "@angular/core";
+
+import * as SuperTokens from "supertokens-website";
+
+SuperTokens.init({
+    apiDomain: "http://localhost:3001",
+    apiBasePath: "/auth",
```
rishabhpoddar, line 7: You can remove apiBasePath

examples/with-angular-thirdpartyemailpassword/README.md
```diff
@@ -0,0 +1,90 @@
...
+-   The frontend code is located in the `/src` folder.
+-   Our frontend will have 3 Angular components, a `home`, `auth` and `root` component.
+-   The `home` component will use `supertokens-website` for session management and the `auth` component will use `supertokens-auth-react` for Authentication flows and UI.
+-   Since both the `home` and `auth` components (since `supertokens-auth-react` SDK uses `supertokens-website`) depend on `supertokens-website`, we can initialize this library in our root component. This will also reduce the bundle sizes for the `home` and `auth` component.
+-   The `root` component contains our routing logic with refrences to the the `home` and `auth` component modules to enable code splitting.
+
+## Code splitting and bundle analysis
+
+Since we use refrences to the `home` and `auth` components in our routing logic, code splitting will be enabled. During the build process Angular will create seperate bundles for the `auth` and `home` components.
```
rishabhpoddar, line 50: typo here

rishabhpoddar (changes requested):

PR #477, opened 2022-06-06
feat: add button to pwless link clicked page if opened on new device

porcellus:
## Summary of change

Adding a button to require user interaction before consuming a pwless link if not opened on the device that created the link.

image

## Related issues

-   

## Test Plan

Updated tests + added a new case that covers opening the link on a separate device.

## Documentation changes

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`


lib/ts/recipe/passwordless/components/features/linkClickedScreen/index.tsx
```diff
@@ -38,15 +38,22 @@ type PropType = FeatureBaseProps & { recipe: Recipe };
...
         if (preAuthSessionId === null || preAuthSessionId.length === 0 || linkCode.length === 0) {
-            return props.recipe.redirectToAuthWithoutRedirectToPath(undefined, props.history, {
+            await props.recipe.redirectToAuthWithoutRedirectToPath(undefined, props.history, {
                 error: "signin",
             });
+            return "REDIRECTING";
+        }
+        const loginAttemptInfo = await props.recipe.recipeImpl.getLoginAttemptInfo({ userContext });
+
+        if (loginAttemptInfo?.preAuthSessionId !== preAuthSessionId) {
```
rishabhpoddar, line 55: if `loginAttemptInfo?.preAuthSessionId === undefined`, then it requires interaction. On the other hand, if `loginAttemptInfo?.preAuthSessionId !== preAuthSessionId`, then don't you want to show `error: "signin"`?

lib/ts/recipe/passwordless/components/features/linkClickedScreen/index.tsx
```diff
@@ -114,6 +126,27 @@ const LinkClickedScreen: React.FC<PropType> = (props) => {
...
+            const preAuthSessionId = getQueryParams("preAuthSessionId");
+            const linkCode = getURLHash();
+
+            if (preAuthSessionId === null || preAuthSessionId.length === 0 || linkCode.length === 0) {
+                // This should never happen, and even if it does the we should be already redirecting
+                throw new Error("Called consumeCode withouth link info");
+            }
+
+            try {
+                const consumeResp = await props.recipe.recipeImpl.consumeCode({
```
rishabhpoddar, line 140: Shouldn't this actually call the modified version of consumeCode you have defined above?

rishabhpoddar (changes requested):

lib/ts/recipe/passwordless/components/features/linkClickedScreen/index.tsx
porcellus: No, that would return `REQUIRES_INTERACTION` again, and the other check is also handled here (without repeating the redirection code). Plus this one updates if `userContext` updates which is not the case for the other.

lib/ts/recipe/passwordless/components/features/linkClickedScreen/index.tsx
porcellus: Not necessarily. Maybe we could add an additional warning to this page, but I don't think we should redirect since logging in through the link is not affected by the stored info. It could indicate, for example, a resent link or some outdated info stored in the current browser if the user wants to continue here after starting the login process elsewhere.

lib/ts/recipe/passwordless/components/themes/translations.ts
```diff
@@ -54,6 +54,10 @@ export const defaultTranslationsPasswordless = {
         PWLESS_SIGN_IN_UP_CHANGE_CONTACT_INFO_EMAIL: "Change email",
         PWLESS_SIGN_IN_UP_CHANGE_CONTACT_INFO_PHONE: "Change phone number",
 
+        PWLESS_LINK_CLICKED_CONTINUE_HEADER: "Sign Up or Log In",
```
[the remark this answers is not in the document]
porcellus: I don't think translation segments should be extracted into constants inside these files, it'd make reading them harder if someone wants to override them, and we really gain much by it.
Reusing the other key is not really that great either because this is a separate screen with entirely separate customizability. I was thinking about making this a bit more specific because it seems disconnected from the other screens where this is the header.

lib/ts/recipe/passwordless/components/features/linkClickedScreen/index.tsx
```diff
@@ -58,8 +65,13 @@ const LinkClickedScreen: React.FC<PropType> = (props) => {
 
     const handleConsumeResp = useCallback(
         async (response: Awaited<ReturnType<typeof consumeCode>>): Promise<void> => {
-            if (!response) {
-                // In this case we are already redirecting
+            if (response === "REQUIRES_INTERACTION") {
+                // We set this here, to make sure it's set after a possible remount
+                setRequiresInteraction(true);
```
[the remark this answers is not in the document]
porcellus: It's false by default, so we may show a spinner while it's loading.

CHANGELOG.md
```diff
@@ -9,7 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
--   Consuming codes on a device different from the
+-   Consuming codes on a device different from the one that created it
```
rishabhpoddar, line 12: Might wanna give more details about this - that we show a button that the user has to click and it's done cause of XYZ reason.

rishabhpoddar (changes requested):

lib/ts/recipe/passwordless/components/features/linkClickedScreen/index.tsx
rishabhpoddar: Can we discuss this in more details: What are all the cases that this if statement would be true in?

lib/ts/recipe/passwordless/components/features/linkClickedScreen/index.tsx
rishabhpoddar: In that case, I think we should rename the `consumeCode` function name on line 41. I was very confused for a while.

lib/ts/recipe/passwordless/components/features/linkClickedScreen/index.tsx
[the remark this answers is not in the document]
rishabhpoddar: But after the user clicks on the button, we want the button to disappear and show the spinner instead right?

lib/ts/recipe/passwordless/components/themes/translations.ts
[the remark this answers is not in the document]
rishabhpoddar: Ok. Let's not use variable and stick to what was done already.

lib/ts/recipe/passwordless/components/features/linkClickedScreen/index.tsx
porcellus: Two groups of cases:
1. Continuing on a new device (w/ no stored info)
2. Stored info about a different attempt

A different attempt could be:
1. Out of date attempt (someone started a login attempt in the same browser)
2. Current attempt from the same user, e.g.: they hit resend but clicked the first email they received, not the last.
3. Current attempt from a diff user - one account started a login process then a link for a different account is clicked

I think in all cases, we are logging in to the expected user and behaving exactly as expected.

rishabhpoddar: Ok agreed.

lib/ts/recipe/passwordless/types.ts
```diff
@@ -264,6 +264,8 @@ export type SignInUpUserInputCodeFormProps = {
 export type LinkClickedScreenProps = {
     recipeImplementation: RecipeInterface;
     config: NormalisedConfig;
+    showButton: boolean;
```
rishabhpoddar, line 267:
```suggestion
    requireUserInteraction: boolean;
```

rishabhpoddar (changes requested):

lib/ts/recipe/passwordless/components/themes/linkClickedScreen/index.tsx
```diff
@@ -17,35 +17,52 @@
...
+                                onClick={() => {
+                                    setLoading(true);
+                                    props.consumeCode();
+                                }}
+                                type="button"
+                                label={"PWLESS_LINK_CLICKED_CONTINUE_BUTTON"}
+                            />
+                        </div>
+                    </React.Fragment>
+                ) : (
```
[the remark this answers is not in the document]
rishabhpoddar: Seems fine

rishabhpoddar (approved):

mufassirkazi:
If you can, please update to these design specs: https://zpl.io/vM5n00Q
Otherwise, its fine

Issue #483, opened 2022-06-09
Updates to the 'Email sent' screen that is shown after reset password form is submitted


st-jay:
## Summary of goal
- Updates as per [this comment](https://github.com/supertokens/main-website/issues/749#issuecomment-1147047323)

> More changes:
> - Updates to the 'Email sent' screen ( which is shown after the reset password form is submitted )
>   - "A password reset email has been sent to <email id>, if it exists in our system."
>   - Button for 'Change or resend email'
>   - Back button linking to the login screen 'Back to log in' ( similar to how it is done on https://test.supertokens.com ) <img width="226" alt="Screenshot 2022-06-06 at 11 15 24 AM" src="https://user-images.githubusercontent.com/92283713/172102283-315e2a45-f373-4a0b-ba3e-4a7949a3d442.png">

PR #485, opened 2022-06-09
feat: Add a link to login form and update copy in reset password email sent screen

st-jay:
## Summary of change

- Added a back to sign in button to the Reset password email sent success screen
- Updated the message displayed on the success screen
- Updated the text of 'Resend' button

## Related issues
-   #483

## Test Plan
-  [video](https://drive.google.com/file/d/1KQLPZOUKFbOmBkV7K29yxwEPyhgcnx9X/view?usp=sharing)

## Documentation changes
None

-   [ ] ~`frontendDriverInterfaceSupported.json` file has been updated (if needed)~
    -   ~Along with the associated array in `lib/ts/version.ts`~
-   [ ] ~Changes to the version if needed~
    -   ~In `package.json`~
    -   ~In `package-lock.json`~
    -   ~In `lib/ts/version.ts`~
-   [ ] ~Issue this PR against the latest non released version branch.~
    -   ~To know which one it is, run find the latest released tag (`git tag`) in the format `vX.Y.Z`, and then find the latest branch (`git branch --all`) whose `X.Y` is greater than the latest released tag.~
    -   ~If no such branch exists, then create one from the latest released branch.~
-   [ ] ~If added a new recipe interface, then make sure that the implementation of it uses NON arrow functions only (like `someFunc: function () {..}`).~
-   [ ] ~If I added a new recipe, I also added the recipe entry point into the `size-limit` section of `package.json` with the size limit set to the current size rounded up.~

None


lib/ts/recipe/emailpassword/components/themes/translations.ts
```diff
@@ -35,9 +35,13 @@ export const defaultTranslationsEmailPassword = {
 
         EMAIL_PASSWORD_RESET_HEADER_TITLE: "Reset your password",
         EMAIL_PASSWORD_RESET_HEADER_SUBTITLE: "We will send you an email to reset your password",
-        EMAIL_PASSWORD_RESET_SEND_SUCCESS: "Please check your email for the password recovery link. ",
-        EMAIL_PASSWORD_RESET_RESEND_LINK: "Resend",
+        EMAIL_PASSWORD_RESET_SEND_SUCCESS_FALLBACK:
```
porcellus, line 38: It's fine as is, but this could've been just a fallback for the "email address" part since it repeats both the before and after email parts.

lib/ts/recipe/emailpassword/components/themes/styles/styles.ts
```diff
@@ -181,6 +181,36 @@ export function getStyles(palette: NormalisedPalette): NormalisedDefaultStyles {
...
+            padding: "0px",
+            cursor: "pointer",
+            lineHeight: "26px",
+            fontSize: "14px",
+            fontWeight: "300",
+            letterSpacing: "0.4px",
+            backgroundColor: "transparent",
+            border: "none",
+            margin: "0px auto",
+            fontFamily: "inherit",
```
porcellus, line 201: There are a few props that seem redundant.
- why do we need to specify `fontFamily: inherit` here?
- `margin` seems to be redundant as well.
- Also, doesn't this match the `secondaryText` style? (and `linkButton`?)

lib/ts/recipe/emailpassword/components/themes/resetPasswordUsingToken/resetPasswordEmail.tsx
```diff
@@ -40,19 +42,36 @@ const EmailPasswordResetPasswordEmail: React.FC<EnterEmailProps> = (props) => {
     };
     const { formFields } = props;
 
+    const getEmailSuccessText = (): string => {
```
porcellus, line 45: This can be a `const` here, no need for a function.

porcellus (changes requested):

lib/ts/recipe/emailpassword/components/themes/styles/styles.ts
porcellus: Maybe reusing `secondaryText` is not the right call here, but please check existing styles to see if any fit here.

st-jay: Without `fontFamily: inherit`, the button element does not update it's font family as per the parent

For instance, the sign in button here does not have `fontFamily: inherit` so it does not use the fontFamily of it's parent
Screenshot 2022-06-14 at 9 08 56 PM

st-jay: `margin` is used to center the button horizontally

porcellus: 1. I didn't know that about `button`s, thanks :)
2. The `display: flex` + related centering props confused me; it seemed either those or the margin should be redundant.

st-jay: It shouldn't matter now, `secondaryText secondaryLinkWithLeftArrow` styles are now used for the button which were already being used for a similar link in Passwordless form.

I have updated it

lib/ts/recipe/emailpassword/types.ts
```diff
@@ -370,6 +370,7 @@ export type EnterEmailProps = FormThemeBaseProps & {
     clearError: () => void;
     onError: (error: string) => void;
     config: NormalisedConfig;
+    onBackButtonClicked: () => void;
```
rishabhpoddar, line 373: why does this PR have the same set of changes as the base PR?

rishabhpoddar (changes requested):

porcellus (approved):
LGTM

rishabhpoddar (approved):

lib/ts/recipe/emailpassword/components/themes/resetPasswordUsingToken/resetPasswordEmail.tsx
```diff
@@ -40,18 +42,29 @@ const EmailPasswordResetPasswordEmail: React.FC<EnterEmailProps> = (props) => {
...
     if (status === "SENT") {
         return (
             <div data-supertokens="container" css={styles.container}>
                 <div data-supertokens="row" css={styles.row}>
                     <div
                         data-supertokens="primaryText enterEmailSuccessMessage"
                         css={[styles.primaryText, styles.enterEmailSuccessMessage]}>
-                        {t("EMAIL_PASSWORD_RESET_SEND_SUCCESS")}
-                        <span data-supertokens="link" css={styles.link} onClick={resend}>
+                        {emailSuccessText}
```
rishabhpoddar, line 59: Need to test:
- Responsiveness on mobile devices
- What happens if you enter a very long email?

rishabhpoddar (changes requested):

rishabhpoddar (approved):

PR #487, opened 2022-06-10
refactor: Refactor cookie handling to only use async functions

nkshah2:
## Summary of change

- Removes usages of the sync version of the cookie handler functions and use async ones instead

## Test Plan

All existing test cases should pass (Github Actions)

## Documentation changes

None

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

-   [ ] Item1
-   [ ] Item2


lib/ts/translation/translationContext.tsx
```diff
@@ -29,27 +30,16 @@ export const TranslationContextProvider: React.FC<
...
-
-            if (res === undefined) {
-                if (fallback !== undefined) {
-                    return fallback;
-                }
-                return key;
-            }
+    const loadLanguageFromCookies = useCallback(
+        async function () {
+            let cookieLang = await getCurrentLanguageFromCookie();
```
porcellus, line 37: I'd prefer if this were a single-line const assignment, but it's fine either way :)

lib/ts/translation/translationContext.tsx
```diff
@@ -65,10 +55,41 @@ export const TranslationContextProvider: React.FC<
         translationControlEventSource.on("LanguageChange", changeHandler);
         translationControlEventSource.on("TranslationLoaded", loadHandler);
 
+        // eslint-disable-next-line @typescript-eslint/no-floating-promises
+        loadLanguageFromCookies();
```
porcellus, line 59:
```suggestion
        void loadLanguageFromCookies();
```

lib/ts/translation/translationContext.tsx
```diff
@@ -29,27 +30,16 @@ export const TranslationContextProvider: React.FC<
...
+    const loadLanguageFromCookies = useCallback(
+        async function () {
+            let cookieLang = await getCurrentLanguageFromCookie();
+            cookieLang = cookieLang === null ? defaultLanguage : cookieLang;
 
-            return res;
+            setCurrentLanguage(cookieLang);
         },
-        [translationStore, currentLanguage, defaultLanguage, userTranslationFunc]
+        [defaultLanguage]
```
porcellus, line 42:
```suggestion
        [defaultLanguage, setCurrentLanguage]
```

lib/ts/translation/translationContext.tsx
```diff
@@ -29,27 +30,16 @@ export const TranslationContextProvider: React.FC<
...
-                }
-                return key;
-            }
+    const loadLanguageFromCookies = useCallback(
+        async function () {
+            let cookieLang = await getCurrentLanguageFromCookie();
+            cookieLang = cookieLang === null ? defaultLanguage : cookieLang;
 
-            return res;
+            setCurrentLanguage(cookieLang);
```
porcellus, line 40:
```suggestion
            setCurrentLanguage((currentVal) => currentVal !== undefined ? currentVal : cookieLang);
```

porcellus (changes requested):

lib/ts/translation/translationContext.tsx
```diff
@@ -8,6 +8,7 @@ import {
     TranslationStore,
 } from "./translationHelpers";
 import { mergeObjects } from "../utils";
+import assert from "assert";
```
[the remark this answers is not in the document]
nkshah2: Done, using the normal throwing "Should never come here" that we use in other places

porcellus (approved):
LGTM

package.json
```diff
@@ -69,7 +69,7 @@
         "react-select": "5.2.1",
         "react-shadow": "19.0.3",
         "supertokens-js-override": "^0.0.4",
-        "supertokens-web-js": "^0.0.2"
+        "supertokens-web-js": "github:supertokens/supertokens-web-js#next-breaking-change"
```
rishabhpoddar, line 72: Keeping this comment until we can set a proper version.

lib/ts/translation/translationContext.tsx
```diff
@@ -29,27 +29,16 @@ export const TranslationContextProvider: React.FC<
...
-                }
-                return key;
-            }
+    const loadLanguageFromCookies = useCallback(
+        async function () {
+            let cookieLang = await getCurrentLanguageFromCookie();
+            cookieLang = cookieLang === null ? defaultLanguage : cookieLang;
 
-            return res;
+            setCurrentLanguage((current) => (current !== undefined ? current : (cookieLang as string)));
```
rishabhpoddar, line 39: add comment explaining why we are keeping current over `cookieLang`

lib/ts/translation/translationContext.tsx
```diff
@@ -29,27 +29,16 @@ export const TranslationContextProvider: React.FC<
...
-            if (res === undefined) {
-                if (fallback !== undefined) {
-                    return fallback;
-                }
-                return key;
-            }
+    const loadLanguageFromCookies = useCallback(
+        async function () {
+            let cookieLang = await getCurrentLanguageFromCookie();
+            cookieLang = cookieLang === null ? defaultLanguage : cookieLang;
```
rishabhpoddar, line 37:
```suggestion
            const cookieLangTemp = cookieLang === null ? defaultLanguage : cookieLang;
```

lib/ts/translation/translationContext.tsx
rishabhpoddar, line 39:
```suggestion
            setCurrentLanguage((current) => (current !== undefined ? current : cookieLangTemp));
```

lib/ts/translation/translationContext.tsx
```diff
@@ -65,10 +54,42 @@ export const TranslationContextProvider: React.FC<
         translationControlEventSource.on("LanguageChange", changeHandler);
         translationControlEventSource.on("TranslationLoaded", loadHandler);
 
+        void loadLanguageFromCookies();
+
         return () => {
             translationControlEventSource.off("LanguageChange", changeHandler);
             translationControlEventSource.off("TranslationLoaded", loadHandler);
         };
     });
```
rishabhpoddar, line 63: shouldn't this run only on mount?

rishabhpoddar (changes requested):

lib/ts/translation/translationContext.tsx
```diff
@@ -29,29 +29,29 @@ export const TranslationContextProvider: React.FC<
...
-            const res = translationStore[currentLanguage] && translationStore[currentLanguage][key];
-            const fallback = translationStore[defaultLanguage] && translationStore[defaultLanguage][key];
-
-            if (res === undefined) {
-                if (fallback !== undefined) {
-                    return fallback;
-                }
-                return key;
-            }
+    const loadLanguageFromCookies = useCallback(
```
porcellus, line 34: No need for `useCallback`, this can be just defined inside the `useEffect`, which could safely use the same dependency array, since this doesn't overwrite anything if the current language is set.

lib/ts/translation/translationContext.tsx
```diff
@@ -70,5 +70,35 @@ export const TranslationContextProvider: React.FC<
...
+                return res;
+            }
+
+            throw new Error("Should never come here");
+        },
+        [translationStore, currentLanguage, defaultLanguage, userTranslationFunc]
+    );
+
+    if (currentLanguage === undefined) {
+        return <></>;
```
porcellus, line 100: I think this should return `null`.

porcellus (changes requested):

lib/ts/translation/translationContext.tsx
nkshah2: Done. We only need this to run on mount so not using any dependencies

porcellus: I think it's better to add the dependency array if it doesn't cause problems. Otherwise, we should start to disable the lint rule specifically, since I think it'd be great to turn it back on.
```
// eslint-disable-next-line react-hooks/exhaustive-deps
```
It's not really an issue, you can resolve this if you like.

nkshah2: Fair enough, added the dependencies on defaultLanguage and setCurrentLanguage. Like you said, it doesnt affect anything so doesnt hurt

rishabhpoddar (approved):

Issue #263, opened 2021-06-12
Make Auth wrapper component work for SSR as well


rishabhpoddar:
- This would be useful in usage for NextJS apps

rishabhpoddar:
## TODO now
The default userContext will be:
```
{
   isLoading: true,
   isDefaultContext: true
}
```

We want to change the session context type to:
```
{
    isLoading: false,
    doesSessionExist: false,
    userId: string,
    accessTokenPayload: {},
} | {
   isLoading: true,
   isDefaultContext: boolean
}
```

However, we would change the return type of `useSessionContext` to be:
```
{
    isLoading: false,
    doesSessionExist: false,
    userId: string,
    accessTokenPayload: {},
} | {
   isLoading: true,
}
```

Because, the user should not care about `isDefaultContext`. We should also throw in `useSessionContext` if it's being called from a component that's not "eventually" a child a `SessionAuth`

- requireAuth: true and `isLoading: true`:
   - We don't wanna render the child, and render null. 
   - The reason for this is that someone may create a child component and not check the context in there since they expect it to load only if a session exists.
   - One issue with this is that it will cause the user to have to check `!context.isLoading` unnecessarily (from a types point of view) - but that's OK. In the future, we can checkout React suspense to solve this.
- else:
   - We render the child with whatever the context state is, which on the server side, will always be `isLoading: true` (since that's the default context).



The reason we have isLoading instead of rendering null is cause maybe many of the SSR pages don't actually care about checking the session context, in which case they would render fine, as opposed to not being rendered (if we were to render null).

## TODO later
### User doesn't care about server side rendering (ssr prop not give):
Same case as what we already have

### User who care about server side rendering (ssr prop give):
- requireAuth: true, userId exists:
   - give the proper session context (session does exist).
- requireAuth: true, userId doesn't exists:
   - We don't wanna render the child, and render null. 
- requireAuth: false, userId exists:
   - give the proper session context (session does exist).
- requireAuth: false, userId doesn't exists:
   - give the proper session context (session does not exist).

## Affect on supertokensWrapper:
- This would itself get the ssr prop and pass it ontp `<SessionAuth requireAuth={false} ssr={...}/>`
- This would always result in the children rendering on server side.

PR #493, opened 2022-06-16
feat: added loading to session state + made auth wrapper to be SSR compatible

porcellus:
## Summary of change

- Updated session context to include loading stat
- Added a session context wrapper devs can use to wrap their entire app
- Made SessionAuth SSR compatible

## Related issues

-   #263 

## Test Plan
- [x] Update existing tests
- [x] Test `requireAuth=false` can render with `loading=true` first

## Documentation changes

- [ ] New wrapper docs
- [ ] Update next.js example

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,8 +17,117 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Breaking changes
 
--   Added `loading` to the session context. Please check if the session context is still loading before using other props. `loading` is always true inside components wrapped by `AuthWrapper`s with `requireAuth=true`
--   Rendering components wrapped by `AuthWrappers` while loading if `requireAuth` is set to false.
+-   Added `loading` to the session context. Please check if the session context is still loading before using other props. `loading` is always false inside components wrapped by `AuthWrapper`s with `requireAuth=true`
+
+### Migration
+
+Following is an example of how your components may have to change. If you had components like this:
+
+```tsx
+// .... other imports
+import SuperTokens, { getSuperTokensRoutesForReactRouterDom, SuperTokensWrapper } from "supertokens-auth-react";
+import { useSessionContext } from "supertokens-auth-react/recipe/session";
+import { ThirdPartyEmailPasswordAuth } from "supertokens-auth-react/recipe/thirdpartyemailpassword";
+
+function Component() {
+    const sessionContext = useSessionContext();
+
+    return <div className="fill">{sessionContext.doesSessionExist ? "logged in" : "logged out"}</div>;
+}
+
+SuperTokens.init({
+    /* config... */
+});
+
+function App() {
+    return (
+        <div className="App">
+            <Router>
+                <div className="fill">
+                    <Routes>
+                        {getSuperTokensRoutesForReactRouterDom(require("react-router-dom"))}
+                        <Route
+                            path="/some-path"
+                            element={
+                                <ThirdPartyEmailPasswordAuth requireAuth={true}>
+                                    {/*
+                                        In this case, Component will always display logged in.
+                                        If doesSessionExist is false after the session is loaded, the user is redirected to the login screen
+                                    */}
+                                    <Component />
+                                </ThirdPartyEmailPasswordAuth>
+                            }
+                        />
+                        <Route
+                            path="/"
+                            element={
+                                <ThirdPartyEmailPasswordAuth requireAuth={false}>
+                                    <Component />
+                                </ThirdPartyEmailPasswordAuth>
+                            }
+                        />
+                    </Routes>
+                </div>
+            </Router>
+        </div>
+    );
+}
+```
+
+It'd look like this after the update:
+
+```tsx
+// .... other imports
+import SuperTokens, { getSuperTokensRoutesForReactRouterDom, SuperTokensWrapper } from "supertokens-auth-react";
+import { useSessionContext } from "supertokens-auth-react/recipe/session";
+import { ThirdPartyEmailPasswordAuth } from "supertokens-auth-react/recipe/thirdpartyemailpassword";
+
+function Component() {
+    const sessionContext = useSessionContext();
+
+    if (sessionContext.loading === true) {
+        // You could display a loading screen here, but session context loading is very fast
+        // so returning null is better in most cases to avoid content popping in and out
+        return null;
+    }
+
+    return <div className="fill">{sessionContext.doesSessionExist ? "logged in" : "logged out"}</div>;
+}
+
+SuperTokens.init({
+    /* config is unchanged */
+});
+
+function App() {
+    return (
+        <div className="App">
+            <SuperTokensWrapper>
+                <Router>
+                    <div className="fill">
+                        <Routes>
+                            {getSuperTokensRoutesForReactRouterDom(require("react-router-dom"))}
+                            <Route
+                                path="/some-path"
+                                element={
+                                    <ThirdPartyEmailPasswordAuth requireAuth={true}>
+                                        {/* 
+                                            In this case, Component will never hit the loading === true branch.
+                                            It will only be rendered if `loading === false && doesSessionExist === true`
+                                            If doesSessionExist is false after the session is loaded, the user is redirected to the login screen
+                                        */}
+                                        <Component />
+                                    </ThirdPartyEmailPasswordAuth>
+                                }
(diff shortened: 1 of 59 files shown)


lib/ts/recipe/emailverification/components/features/emailVerification/index.tsx
```diff
@@ -51,15 +51,15 @@ export const EmailVerification: React.FC<Prop> = (props) => {
     const fetchIsEmailVerified = useCallback(async () => {
         const token = getQueryParams("token") ?? undefined;
         if (token === undefined) {
-            if (!sessionContext.doesSessionExist) {
+            if (sessionContext.loading === false && !sessionContext.doesSessionExist) {
                 await props.recipe.config.redirectToSignIn(props.history);
             } else {
                 // we check if the email is already verified, and if it is, then we redirect the user
                 return (await props.recipe.recipeImpl.isEmailVerified({ userContext })).isVerified;
             }
```
rishabhpoddar, line 59: Shouldn't we do this entire thing only if `sessionContext.loading === false`? As opposed to doing `return (await props.recipe.recipeImpl.isEmailVerified({ userContext })).isVerified;` if `sessionContext.loading === true`

rishabhpoddar (changes requested):

lib/ts/recipe/session/index.ts
```diff
@@ -76,6 +78,7 @@ const signOut = SessionAPIWrapper.signOut;
 export {
     useSessionContext,
     SessionAuth,
+    SuperTokensWrapper,
```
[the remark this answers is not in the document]
porcellus: I'm not sure: this doesn't work without the session recipe, so it feels appropriate here. Maybe we could make it work (or not throw) without an initialized session. Maybe just a warn?

lib/ts/recipe/emailverification/components/features/emailVerification/index.tsx
porcellus: You are correct, but this will only ever run if `loading` is false.  Let's discuss this along with `useOnMountAPICall` on a call tomorrow.

lib/ts/recipe/emailverification/components/features/emailVerification/index.tsx
```diff
@@ -51,15 +51,15 @@ export const EmailVerification: React.FC<Prop> = (props) => {
...
         }
         return false;
-    }, [props.recipe]);
+    }, [props.recipe, sessionContext]);
```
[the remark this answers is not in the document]
porcellus: It'd have the same effect (after `sessionContext.loading` is set to true, this is called and the result is saved) but it'd make the linter complain.

CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [unreleased]
 
+### Changes
+
+-   Added `loading` to the session context
+-   Added `SuperTokensWrapper` intended to wrap around whole applications, providing a session context
+-   Made auth wrappers SSR compatible
```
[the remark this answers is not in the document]
porcellus: Right, I'll move these into the Breaking changes section. Where/how should I add the example?

lib/ts/recipe/emailverification/components/features/emailVerification/index.tsx
porcellus: I'll add an assert instead above and a comment explaining.

lib/ts/recipe/emailverification/emailVerificationAuth.tsx
```diff
@@ -29,45 +29,44 @@ const EmailVerificationAuth: React.FC<Props> = ({ children, ...props }) => {
...
         },
-        [doesSessionExist, emailVerificationMode]
+        [sessionContext.loading, emailVerificationMode]
     );
 
     useOnMountAPICall(checkIsEmailVerified, useIsEmailVerified, undefined, sessionContext.loading === false);
 
-    if (doesSessionExist === false) {
-        return <>{children}</>;
-    }
```
rishabhpoddar, line 64: this should not be removed IMO cause if requireAuth is false, and a session does not exist, it should show the child

rishabhpoddar: Maybe add a test about this

rishabhpoddar (changes requested):

rishabhpoddar (approved):

lib/ts/recipe/emailverification/emailVerificationAuth.tsx
```diff
@@ -29,45 +29,49 @@ const EmailVerificationAuth: React.FC<Props> = ({ children, ...props }) => {
...
-    useOnMountAPICall(checkIsEmailVerified, useIsEmailVerified);
-
-    if (sessionContext.doesSessionExist === false) {
-        return <>{children}</>;
-    }
+    useOnMountAPICall(checkIsEmailVerified, useIsEmailVerified, undefined, sessionContext.loading === false);
 
-    if (props.recipe.config.mode !== "REQUIRED") {
+    if (
+        sessionContext.loading === true ||
```
rishabhpoddar, line 66: are you sure you want to render the children when loading is true? Cause then it might cause a flicker where you see a child, and then loading is false, but the email is not verified, in which case it will redirect.

rishabhpoddar (changes requested):

lib/ts/recipe/emailverification/emailVerificationAuth.tsx
porcellus: We've agreed to change it to render null when loading

lib/ts/recipe/session/sessionAuth.tsx
```diff
@@ -93,22 +96,21 @@ const SessionAuth: React.FC<PropsWithChildren<Props>> = ({ children, ...props })
...
+            loading: false,
         };
     }, []);
 
     const setInitialContextAndMaybeRedirect = useCallback(
         async (toSetContext: SessionContextType) => {
             // if this component is unmounting, or the context has already
             // been set, then we don't need to proceed...
-            if (!toSetContext.doesSessionExist && props.requireAuth === true) {
+            if (toSetContext.loading === false && !toSetContext.doesSessionExist && props.requireAuth === true) {
```
rishabhpoddar, line 107: remove this false check and make it into an assert false thing

lib/ts/recipe/emailverification/emailVerificationAuth.tsx
```diff
@@ -22,52 +22,70 @@ import { useUserContext } from "../../usercontext";
...
-        [doesSessionExist, emailVerificationMode]
+        [sessionContext.loading]
     );
 
-    useOnMountAPICall(checkIsEmailVerified, useIsEmailVerified);
-
-    if (sessionContext.doesSessionExist === false) {
-        return <>{children}</>;
-    }
+    useOnMountAPICall(checkIsEmailVerified, useIsEmailVerified, undefined, sessionContext.loading === false);
```
rishabhpoddar, line 77: Add a comment explaining why we have broken down `checkIsEmailVerified`, `useIsEmailVerified`.

lib/ts/recipe/emailverification/emailVerificationAuth.tsx
```diff
@@ -22,52 +22,70 @@ import { useUserContext } from "../../usercontext";
...
 
-    if (props.recipe.config.mode !== "REQUIRED") {
+    const isNotRequired = recipe.current !== undefined && recipe.current.config.mode !== "REQUIRED";
+    if (
+        // We only render after loading has finished to eliminate flicker
+        sessionContext.loading === false &&
+        (isNotRequired || sessionContext.doesSessionExist === false || isEmailVerified)
+    ) {
```
rishabhpoddar, line 84: Please split this up into more if statements to make it more clear

rishabhpoddar (changes requested):

lib/ts/recipe/session/sessionAuth.tsx
rishabhpoddar: @porcellus this false check is not needed anymore.

rishabhpoddar (changes requested):

rishabhpoddar (approved):

PR #515, opened 2022-07-04
docs: updated examples for new session context type

porcellus:
## Summary of change

Updated examples for the new session context type

## Related issues

-   #493 

## Test Plan

N/A, example update

## Documentation changes

N/A, example update

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

### DIFF
--- a/examples/with-thirdpartypasswordless/src/App.js
+++ b/examples/with-thirdpartypasswordless/src/App.js
@@ -1,6 +1,6 @@
 import { useState } from "react";
 import "./App.css";
-import SuperTokens, { getSuperTokensRoutesForReactRouterDom } from "supertokens-auth-react";
+import SuperTokens, { SuperTokensWrapper, getSuperTokensRoutesForReactRouterDom } from "supertokens-auth-react";
 import ThirdPartyPasswordless from "supertokens-auth-react/recipe/thirdpartypasswordless";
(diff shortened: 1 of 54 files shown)


examples/with-thirdpartypasswordless/src/App.js
```diff
@@ -1,6 +1,6 @@
 import { useState } from "react";
 import "./App.css";
-import SuperTokens, { getSuperTokensRoutesForReactRouterDom } from "supertokens-auth-react";
+import SuperTokens, { SuperTokensWrapper, getSuperTokensRoutesForReactRouterDom } from "supertokens-auth-react";
```
rishabhpoddar, line 3: - Need to update package.json in all apps too to use the latest version of SDK.
- Have you tested all the apps in which you have made changes to?

porcellus: - I thought that should be done when it was getting released. Should I do it now?
- I just ran through them with a few exceptions (supabase, AWS, hasura, netlify, with-next-iframe). There are a few cases where I ran into issues, but they are unrelated to this PR. I'll add an issue to test, check and fix the example apps.

porcellus: #516

PR #521, opened 2022-07-08
feat(sessionclaims): add&use email verification claim

porcellus:
## Summary of change

Add email verification claim

## Related issues

-  https://github.com/supertokens/supertokens-node/pull/349
-  https://github.com/supertokens/supertokens-web-js/pull/48

## Test Plan

Done separately

## Documentation changes

Done separately

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`


lib/ts/recipe/emailverification/recipe.tsx
```diff
@@ -67,6 +69,14 @@ export default class EmailVerification extends RecipeModule<
             );
             this.recipeImpl = builder.override(this.config.override.functions).build();
         }
+        if (this.config.mode !== "OFF") {
+            SessionClaimValidatorStore.addClaimValidatorFromOtherRecipe(
```
rishabhpoddar, line 73: We should add a test which enforces that the order of doing recipe init in the recipe list affects the priority of the redirection in case of claim validation failure.

rishabhpoddar (changes requested):

lib/ts/recipe/emailverification/recipe.tsx
porcellus: Will be done in a separate PR

lib/ts/components/supertokensWrapper.tsx
```diff
@@ -1,17 +1,10 @@
...
+    PropsWithChildren<{
+        userContext?: any;
+    }>
 > = (props) => {
-    return (
-        <UserContextWrapper userContext={props.userContext}>
-            <SessionAuthWrapper {...props} />
-        </UserContextWrapper>
-    );
+    return <SessionAuthWrapper {...props} requireAuth={false} overrideGlobalClaimValidators={() => []} />;
```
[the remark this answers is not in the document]
porcellus: I think it's better if it's fully explicit here. Also, we've discussed making that change, so I'll add it here.

rishabhpoddar (approved):

PR #551, opened 2022-08-04
Feat: expose no redirect auth comps

porcellus:
## Summary of change

Expose `NoRedirect` version of auth components that do not redirect away if there is an active session.

## Related issues

- 

## Test Plan

These components are already tested.

## Documentation changes

Not strictly necessary, but we could mention them in the embedding docs.

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

-   [ ] Check if we need docs update
-   [ ] Think about naming (I'm not 100% satisfied with `SignInUpNoRedirect`)


lib/ts/recipe/thirdparty/index.ts
```diff
@@ -172,6 +172,8 @@ export default class Wrapper {
     static Facebook = Facebook;
     static Github = Github;
     static SignInAndUp = (prop?: any) => ThirdParty.getInstanceOrThrow().getFeatureComponent("signinup", prop);
+    static SignInAndUpNoRedirect = (prop?: any) =>
+        ThirdParty.getInstanceOrThrow().getFeatureComponent("signinup-noredirect", prop);
```
rishabhpoddar, line 176: Should we just not exposed a redirect version of that at all? Will be easier to do docs.. otherwise we have to explain somewhere that there are two versions.

rishabhpoddar (changes requested):

lib/ts/recipe/thirdparty/index.ts
porcellus: I think the main use case for that is if devs just want to change the path and/or wrap it in some kind of layout. This seems a bit more common than the no-redirect one. Still, it'd be easier to document.
Another option is to expose them as a single version but add an extra prop that controls if a redirect should happen.

lib/ts/recipe/emailpassword/index.ts
```diff
@@ -167,6 +167,8 @@ export default class Wrapper {
     }
 
     static SignInAndUp = (prop?: any) => EmailPassword.getInstanceOrThrow().getFeatureComponent("signinup", prop);
+    static SignInAndUpNoRedirect = (prop?: any) =>
+        EmailPassword.getInstanceOrThrow().getFeatureComponent("signinup-noredirect", prop);
```
rishabhpoddar, line 171:
```suggestion
    static SignInAndUp = (prop?: {redirectOnSessionExists: boolean = true}) => EmailPassword.getInstanceOrThrow().getFeatureComponent("signinup", prop);
```

rishabhpoddar (changes requested):

rishabhpoddar (changes requested):

examples/with-thirdpartyemailpassword-2fa-passwordless/src/SecondFactor/index.tsx
```diff
@@ -30,7 +30,7 @@ export default function SecondFactor() {
                 alignItems: "center",
                 justifyContent: "center",
             }}>
-            <Passwordless.SignInUp />
+            <Passwordless.SignInUp redirectOnSessionExists={false} />
```
rishabhpoddar, line 33: this hasn’t made the example apps simpler.. so what is the benefit of doing this? Also, it has introduced a bug in which if you are logged in, and you visit the second factor screen, it keeps you on that screen (it sometimes navigates you away, but it glitches).

rishabhpoddar (changes requested):

PR #558, opened 2022-08-12
feat: session auth refactor

porcellus:
## Summary of change

- Cleaned up some types of the session context
- Removed `redirectToLogin` from `SessionAuth` (it was in the changelog and unused but not removed, this change was lost somewhere)
- Added `useClaimValue` + added it into a test

## Related issues

-   

## Test Plan

Updated a test that includes it.

## Documentation changes

Done separately

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`


lib/ts/recipe/session/sessionContext.ts
```diff
@@ -1,14 +1,9 @@
 import React from "react";
 import { SessionContextType } from "./types";
 
-const SessionContext = React.createContext<SessionContextType & { isDefault: boolean }>({
+const SessionContext = React.createContext<SessionContextType & { isDefault?: boolean }>({
```
rishabhpoddar, line 4: why is isDefault optional? I thought that it will be used by SessionAuth to know that if it's already enclosed by a parent SessionAuth or not, and also that the user exposed context type will not have this. So for internal use, shouldn't we keep isDefault compulsory?

lib/ts/recipe/session/sessionAuth.tsx
```diff
@@ -184,9 +179,7 @@ const SessionAuth: React.FC<PropsWithChildren<SessionAuthProps>> = ({ children,
         return null;
     }
 
-    return (
-        <SessionContext.Provider value={{ ...context, isDefault: false } as any}>{children}</SessionContext.Provider>
-    );
+    return <SessionContext.Provider value={context}>{children}</SessionContext.Provider>;
```
rishabhpoddar, line 182: why are we not giving isDefault as false anymore?

lib/ts/recipe/session/useSessionContext.ts
```diff
@@ -6,7 +6,7 @@ import { SessionContextType } from "./types";
 const useSessionContext = (): SessionContextType => {
     const ctx = React.useContext(SessionContext);
 
-    if (ctx.isDefault) {
+    if (ctx.isDefault === true) {
         throw new Error("Cannot use useSessionContext outside auth wrapper components.");
     }
```
rishabhpoddar, line 11: why this change? Cause not if isDefault is undefined, it will not throw the error - and from the types, it can be undefined. When will it be undefined?

rishabhpoddar (changes requested):

lib/ts/recipe/session/sessionAuth.tsx
porcellus: We are checking for strict equality anyway and this way we don't need the `as any` to make the typing work.

lib/ts/recipe/session/sessionContext.ts
porcellus: - it makes typing it properly in `SessionAuth` easier (it could be done I think, but more work)
- we are checking strict equality anyway.

lib/ts/recipe/session/useSessionContext.ts
porcellus: It's undefined if it's provided by `SessionAuth` (after the change). We only need that prop in the default/initialization value of the context. We could set it to false everywhere else but just omitting it is fine since we never actually want to set it to true.

rishabhpoddar (approved):

PR #568, opened 2022-08-22
feat: add doRedirection to SessionAuth

porcellus:
## Summary of change

Adding `doRedirection` to `SessionAuth` as discussed. 
It defaults to true, controlling if invalid claims and missing session/the UNAUTHORISED event should redirect.

## Related issues

-   

## Test Plan

Added unit tests.

## Documentation changes

N/A

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

### DIFF
--- a/lib/ts/recipe/session/sessionAuth.tsx
+++ b/lib/ts/recipe/session/sessionAuth.tsx
@@ -28,6 +28,7 @@ import { SessionClaimValidator } from "supertokens-website";
 
 export type SessionAuthProps = {
     requireAuth?: boolean;
+    doRedirection?: boolean;
     onSessionExpired?: () => void;
     overrideGlobalClaimValidators?: (
         globalClaimValidators: SessionClaimValidator[],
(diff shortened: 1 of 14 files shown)


rishabhpoddar:
Let's discuss this on call. Either im confused about something, or this prop, the way it is now, doesn't really make sense

lib/ts/recipe/session/sessionAuth.tsx
```diff
@@ -28,6 +28,7 @@ import { SessionClaimValidator } from "supertokens-website";
 
 export type SessionAuthProps = {
     requireAuth?: boolean;
+    doRedirection?: boolean;
```
rishabhpoddar, line 31: Add a comment for the user like -> "To know the behaviour of requireAuth, please visit ..."

rishabhpoddar (changes requested):

lib/ts/recipe/session/sessionAuth.tsx
porcellus: image

rishabhpoddar (approved):

PR #569, opened 2022-08-22
fix: session claims 2fa example

porcellus:
## Summary of change

Re-add removing the change phone number footer to the 2fa example.

## Related issues

-   

## Test Plan

N/A example only test

## Documentation changes

N/A example only test

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

### DIFF
--- a/examples/with-thirdpartyemailpassword-2fa-passwordless/src/App.tsx
+++ b/examples/with-thirdpartyemailpassword-2fa-passwordless/src/App.tsx
@@ -6,7 +6,7 @@ import SuperTokens, {
     redirectToAuth,
 } from "supertokens-auth-react";
 import ThirdPartyEmailPassword from "supertokens-auth-react/recipe/thirdpartyemailpassword";
-import Session, { SessionAuth } from "supertokens-auth-react/recipe/session";
+import Session, { SessionAuth, useSessionContext } from "supertokens-auth-react/recipe/session";
 import Home from "./Home";
 import { Routes, BrowserRouter as Router, Route, useLocation } from "react-router-dom";
 import Footer from "./Footer";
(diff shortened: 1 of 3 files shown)


examples/with-thirdpartyemailpassword-2fa-passwordless/src/App.tsx
```diff
@@ -78,6 +78,37 @@ SuperTokens.init({
                             </div>
                         );
                     },
+                    // we override the component which shows the change phone number button
+                    PasswordlessUserInputCodeFormFooter_Override: ({ DefaultComponent, ...props }) => {
+                        // the logic here is very similar to the previous step's logic
```
rishabhpoddar, line 83: which previous step?

examples/with-thirdpartyemailpassword-2fa-passwordless/src/App.tsx
```diff
@@ -78,6 +78,37 @@ SuperTokens.init({
...
+                                    }
+                                })
+                                .catch((err) => {
+                                    // it can come here if a session doesn't exist.
+                                    // in this case, the screen we should redirect to the
+                                    // first login challenge
+                                    SuperTokens.redirectToAuth({
+                                        redirectBack: false,
+                                    });
+                                });
```
rishabhpoddar, line 102: You can change this to call doesSessionExist and if it doesn't call 
```
SuperTokens.redirectToAuth({
                                        redirectBack: false,
                                    });
```

else do the accessTokenPayload thing.

The reason it was done like this previously was cause we had overriden doesSessionExist to return false even if it did exist and so the easiest way to know if it actually existed was by calling accessTokenPayload and it throwing an error.

rishabhpoddar (changes requested):

rishabhpoddar (approved):

PR #573, opened 2022-08-24
fix: update test examples action & fix errors

porcellus:
## Summary of change

- Updated size limits
- Updated `test-examples` action to run on all pushes
- Updated `test-examples` action to install supertokens-auth from current commit

## Related issues

-   

## Test Plan

N/A ci update

## Documentation changes

N/A ci update

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`


rishabhpoddar (approved):

porcellus:
The changes to session auth were necessary to solve an edge case (reliably occurring in tests, but not on my PC), where:
## Edge case 1.
On page load:
1. `SuperTokensWrapper` starts validation for all global claim validators (incl. email verification)
2. During the validation process above, the `SessionAuth` also starts validation/session loading.
3. SuperTokensWrapper validation, during which the email verification validator modifies the user context to request redirection to the email verification screen.
4. SessionAuth validation also finishes, sees the modified user context, and tries to redirect to the email verification screen even though it's not supposed to check the email verification claim.
This was solved by:
- removing the redirection request from the user context even if the current `SessionAuth` is not doing redirects.
- moving the above removal closer to the end of the validation call to minimize the timing where this could occur.
This should now be very rare, but it could occur if devs add async validators that execute for a long time.

## Edge case 2.
We could get an `UNAUTHORISED` event before the session context is fully loaded. This is not a big issue, but it could cause a strange issue. It'd call the `onSessionExpired` (and setting `doesSessionExist: false`) and/or redirect the user away while the context is still building. 
This doesn't usually cause issues, but by simply adding the event handler after the session loads solves this.

Issue #453, opened 2022-05-18
Overriding default provider's functions causes issues with styling


rishabhpoddar:
If we do the following:
```tsx
let google = ThirdPartyEmailPassword.Google.init();

ThirdPartyEmailPassword.init({
    signInAndUpFeature: {
        providers: [{
            ...google,
            getRedirectURL: () => {
                return "https://auth.abc.cloud/auth/callback/google"
            }
        }]
    }
})
```

The rendered button for google no longer has the google branding.

rishabhpoddar:
Fixed in version >= 0.25.0

PR #576, opened 2022-09-07
feat: make getRedirectURL configurable in third-party providers

porcellus:
## Summary of change

Added a way to override `getRedirectURL`; This makes all methods configurable besides `generateState`, which I think shouldn't be.

## Related issues

-   #453

## Test Plan

Manual test (automated test requires updating our config on auth0)

## Documentation changes

Not strictly necessary; this is very likely a rarely-used prop

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`


lib/ts/recipe/thirdparty/providers/index.tsx
```diff
@@ -52,7 +55,7 @@ export default abstract class Provider {
         return <ProviderButton logo={this.getLogo()} providerName={providerName} displayName={this.name} />;
     }
 
-    getRedirectURL(): string {
+    defaultGetRedirectURL(): string {
         const domain = SuperTokens.getInstanceOrThrow().appInfo.websiteDomain.getAsStringDangerous();
```
rishabhpoddar, line 59: Will the routing logic need to change for the SDK to now detect these new paths here: https://github.com/supertokens/supertokens-auth-react/blob/63aa6a289d/lib/ts/recipe/thirdparty/recipe.tsx#L95

lib/ts/recipe/thirdparty/providers/types.ts
```diff
@@ -33,6 +34,12 @@ export type BuiltInProviderConfig = {
     buttonComponent?: JSX.Element;
```
rishabhpoddar, line 35: no tests for this change?

rishabhpoddar (changes requested):

lib/ts/recipe/thirdparty/providers/types.ts
porcellus: We could add one, but we need to update the auth0 config to allow the test redirection url

lib/ts/recipe/thirdparty/providers/index.tsx
porcellus: I think that should be handled by embedding the feature component, as this could be any path (outside of `websiteBasePath` too), and it could even be on an entirely different sub-domain (although that'd require more customization).

rishabhpoddar: hmm. okay.

lib/ts/recipe/thirdparty/providers/types.ts
rishabhpoddar: Let's discuss on call

rishabhpoddar (approved):

rishabhpoddar (approved):

Issue #580, opened 2022-09-09
app in broken state if supertokens user got deleted and email verification is enabled


n1ru4l:
I spotted some kind of bug on the front end that leaves the application in an unusable state as the HTTP call to `/api/auth/user/email/verify` raises the error `Error: Unknown User ID provided (and makes the app unusable)`. 

Full Stack on the backend of the original error:

```
Error: Unknown User ID provided
    at Recipe.<anonymous> (/app/node_modules/supertokens-node/lib/build/recipe/thirdpartyemailpassword/recipe.js:129:27)
    at Generator.next (<anonymous>)
    at fulfilled (/app/node_modules/supertokens-node/lib/build/recipe/thirdpartyemailpassword/recipe.js:15:36)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

I think this is related due to an account that was deleted but the `sAccessToken`, `sFrontToken`, `sIRTFrontend` and `sIdRefreshToken` cookies still are present on the frontend.

After deleting the cookies from the dev tools everything works fine!

It seems like there should be better handling of this edge case instead of instructing users to clear their cookies. 🤔 

This might be related to https://github.com/supertokens/supertokens-auth-react/issues/547 ?

porcellus:
Thanks for filing this as an issue on GH; it's easier to track :)
I think it's unrelated to think linked issue. This should error should not get to the application and be handled inside the SDK.
We should trigger a session refresh on the frontend (so basically return a 401 as we do for an expired access token), because getting this error means that the user has an invalid session.

rishabhpoddar:
This issue has been solved now. The email verification APIs will return a 401 in case of a deleted user which will redirect the user to the auth screen.

PR #588, opened 2022-09-16
fix: handling active sessions of deleted users

porcellus:
## Summary of change

-   Redirecting to auth when opening the default UI on `/auth/verify-email` with an unknown/deleted user instead of showing an empty screen
-   In case an error is thrown during loading session information/claim validation, `sessionContext` will now contain `doesSessionExist: false` instead of throwing an error

## Related issues

-   #580 

## Test Plan

Added E2E tests

## Documentation changes

N/A, this is a bugfix mainly

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`


lib/ts/recipe/emailverification/components/themes/emailVerification/sendVerifyEmail.tsx
```diff
@@ -86,7 +88,11 @@ export const EmailVerificationSendVerifyEmail: React.FC<SendVerifyEmailThemeProp
         [props.config, props.recipeImplementation, props.onEmailAlreadyVerified]
     );
 
-    const handleSendError = useCallback(() => {
+    const handleSendError = useCallback(async () => {
+        // If the error cleared the session we should redirect to auth
+        if ((await Session.getInstanceOrThrow().doesSessionExist({ userContext })) !== true) {
+            await props.redirectToAuth();
+        }
```
rishabhpoddar, line 95: Add a TODO stating that when we add SessionAuth with requireAuth=true, we should remove this.

lib/ts/recipe/emailverification/components/features/emailVerification/index.tsx
```diff
@@ -87,7 +88,24 @@ export const EmailVerification: React.FC<Prop> = (props) => {
...
+        async (err) => {
+            // If the error cleared the session we redirect away, otherwise we have no way of handling it.
+            if (await Session.getInstanceOrThrow().doesSessionExist({ userContext })) {
+                throw err;
+            } else {
+                await redirectToAuthWithHistory();
+            }
+        },
+        [redirectToAuthWithHistory]
+    );
```
rishabhpoddar, line 106: Add a TODO stating that when we add SessionAuth with requireAuth=true, we should remove this.

lib/ts/recipe/session/sessionAuth.tsx
```diff
@@ -98,25 +98,53 @@ const SessionAuth: React.FC<PropsWithChildren<SessionAuthProps>> = ({ children,
...
+            // Which means that the session was actually invalid meaning returning false is right.
+            // This might also happen if the user provides an override or a custom claim validator that throws (or if we have a bug)
+            // In one case it's the right way to handle this, and the other is a user error (and it should be rare)
+            return {
+                loading: false,
+                doesSessionExist: false,
+                accessTokenPayload: {},
+                invalidClaims: [],
+                userId: "",
+            };
```
rishabhpoddar, line 119: if a session exists, then we should throw ex, else we should return doesSessionExist false

lib/ts/recipe/session/sessionAuth.tsx
```diff
@@ -98,25 +98,53 @@ const SessionAuth: React.FC<PropsWithChildren<SessionAuthProps>> = ({ children,
...
+        } catch {
+            // This means that loading the access token or the userId failed
+            // This may happen if the server cleared the error since the validation was done which should be extremely rare
+            return {
+                loading: false,
+                doesSessionExist: false,
+                accessTokenPayload: {},
+                invalidClaims: [],
+                userId: "",
+            };
```
rishabhpoddar, line 146: if a session exists, then we should throw ex, else we should return doesSessionExist false

rishabhpoddar (changes requested):

rishabhpoddar (approved):

Issue #522, opened 2022-07-10
@emotion/react & @emotion/cache must be installed separately


doei:
When setting up supertokens I got this error:
`supertokens Module not found: react-shadow tried to access @emotion/cache, but it isn't declared in its dependencie`

After instally @emotion/cache, I had the same error with @emotion/react

Installing them both seem to solve the problem, there might be an issue with react-shadow

rishabhpoddar:
Thanks @doei . We will get back ASAP.

rishabhpoddar:
@doei can we have a look at the package-lock.json please? This may be happening because there might exist some other dependency in your app that also depends on react-shadow (thereby promoting it to the top level node_modules), and even though react-shadow does not have emotion in it's dependencies, it uses it. Hence this error.

JakeGinnivan:
Emotion should be a peer dependency IMO. That way it works in nicely with other libraries like ChakraUI

See https://github.com/chakra-ui/chakra-ui/blob/main/packages/react/package.json#L65

doei:
This is my package.json, however, I'm in the process of completely upgrading this project, so hopefully the issue will solve itself. 

```json
{
	"version": "0.1.0",
	"private": true,
	"dependencies": {
		"@emotion/cache": "^11.9.3",
		"@emotion/react": "^11.9.3",
		"@sentry/react": "^6.1.0",
		"@testing-library/jest-dom": "^5.11.4",
		"@testing-library/react": "^11.1.0",
		"@testing-library/user-event": "^12.1.10",
		"date-fns": "^2.17.0",
		"debounce": "^1.2.1",
		"markdown-it": "^12.0.4",
		"prop-types": "^15.7.2",
		"react": "^17.0.1",
		"react-day-picker": "^8.0.0-beta.3",
		"react-dom": "^17.0.1",
		"react-is": "^17.0.2",
		"react-loader-spinner": "^4.0.0",
		"react-markdown": "^5.0.3",
		"react-scripts": "4.0.2",
		"react-spring": "^9.1.1",
		"recoil": "^0.1.2",
		"supertokens-auth-react": "^0.24.0",
		"supertokens-website": "^13.0.1",
		"web-vitals": "^1.0.1"
	},
	"scripts": {
		"start": "npx npm-run-all -p watch-css start-js",
		"start-js": "BROWSER=none PORT=8080 react-scripts start",
		"watch-css": "npm run build-css && stylus -w ./src/style.styl -o ./src/style.css",
[...]

PR #610, opened 2022-11-03
Include implicit deps in package.json

enisdenjo:
supertokens-auth-react [imports supertokens-website](https://github.com/supertokens/supertokens-auth-react/blob/e1995f65c08690586b802218efdd4876183aa87a/lib/ts/utils.ts#L18) (through supertokens-web-js), and [imports @emotion/cache](https://github.com/supertokens/supertokens-auth-react/blob/bfefc2983b0d21f6f3b0f2539fb66e0a88973b62/lib/ts/components/featureWrapper.tsx#L24) (through react-select); while none of these dependencies appear in [its package.json](https://github.com/supertokens/supertokens-auth-react/blob/bfefc2983b0d21f6f3b0f2539fb66e0a88973b62/package.json#L63-L73).

Implicit dependencies are disallowed and simply will not work with [pnpm](https://github.com/pnpm/pnpm).

Causes https://github.com/kamilkisiela/graphql-hive/pull/590.

P.S. I had to [update the lockfile](https://github.com/supertokens/supertokens-auth-react/commit/134bd37821ca7118ff49e90c7e49a6e91d01b289) because it was outdated apparently; and I have to use `--legacy-peer-deps` flag while installing because just `npm i` wont work due to some resolution conficts.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,6 +9,7 @@
             "version": "0.27.1",
             "license": "Apache-2.0",
             "dependencies": {
+                "@emotion/cache": "^11.4.0",
                 "@emotion/react": "^11.4.1",
                 "chroma-js": "2.1.2",
                 "libphonenumber-js": "1.9.43",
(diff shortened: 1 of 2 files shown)


porcellus:
I've reviewed the offending imports:
- We are looking to drop emotion soon
- Direct imports of `supertokens-website` should be fixed instead of adding the dependency
So: Since this is solvable outside the SDK and we are looking to fix both issues in different ways I'll close this PR and:
- open an issue about re-exporting them in supertokens-web-js and importing them through that package.
- you can track #522 about dropping emotion (this should also solve the legacy-peer-deps issue)

PR #624, opened 2022-12-09
feat!: remove emotion

porcellus:
## Summary of change

- Removes emotion, react-select, react-phone-number-input dependencies
- Moves to bundling with rollup

## Related issues

-   

## Test Plan

- Visual testing

## Documentation changes

TODO

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`


porcellus:
We need to update supertokens-website before this can be merged (https://github.com/supertokens/supertokens-website/pull/178)

other/checkTranslationKeys.js
```diff
@@ -11,6 +11,8 @@ async function* walk(rootPath) {
 }
 
 async function checkTranslationKeys() {
+    // TODO: re-implement this
+    return;
```
rishabhpoddar, line 15: TODO - why is this affected? And should we fix this now or later? Cause later it won't end up happening

rishabhpoddar (changes requested):

other/checkTranslationKeys.js
porcellus: This relied on having a separate translation.js file. I'm not 100% sure if we need this. It was mostly for the translation rework; I just left it in place because it was working. 
I could remove this and add a check into the PR template. We can do the re-implementation if we start having problems with this.

rishabhpoddar: See if you can get this to work in this PR itself.

package.json
```diff
@@ -79,12 +80,13 @@
...
-        "supertokens-js-override": "^0.0.4",
-        "supertokens-web-js": "^0.3.0"
+        "supertokens-js-override": "^0.0.4"
     },
     "peerDependencies": {
         "react": ">=16.8.0",
-        "react-dom": ">=16.8.0"
+        "react-dom": ">=16.8.0",
+        "supertokens-web-js": "^0.3.0",
+        "supertokens-website": "supertokens/supertokens-website#feat/getWindowUnsafe"
```
rishabhpoddar, line 89: why do you need to add supertokens-website to the peer dependency? Shouldn't web-js already have that as a dependency?

rishabhpoddar (changes requested):

package.json
porcellus: web-js has the dependency, but we also import it directly here. while I think this should be fixed, right now, it'd result in the website SDK being bundled. If you prefer, I can remove this and add the website SDK into the bundler config as an external dependency, but it shouldn't be bundled.

rishabhpoddar (approved):

rishabhpoddar (approved):

Issue #625, opened 2022-12-09
Wrong back button link


insky:
If I click "Forgot password" and then back button, URL is {websiteDomain}/{websiteBasePath}?show=signin (no trailing slash), but {websiteDomain}/{websiteBasePath}/ (with trailing slash) expected.

rishabhpoddar:
hey @insky why does having a trailing slash (or not) matter?

insky:
@rishabhpoddar because of uniformity of routes. For example, we have {websiteDomain}/{websiteBasePath}/reset-password
Anyway, component doesn't consider have websiteBasePath trailing slash or not.

rishabhpoddar:
Yea.. I don't think this matters too much. Im closing this issue.

egorFiNE:
@rishabhpoddar we have the same issue. Some http routing configurations are unexpectedly broken because of that. We have routed everything `/{websiteBasePath}/` to the backend and then found out that reset password is broken due to the fact that supertokens routes this one specific request to a _different_ albeit very similar url. 

So we had to add additional deciphering logic. 

Before: 

* does the URL begin with `/{websiteBasePath}/`? Yes → route to backend

After we have discovered this issue: 

* does the URL begin with `/{websiteBasePath}/`? Yes → route to backend
* does the URL begin with `/{websiteBasePath}`? 
  * is this the full url? Yes → route to backend
  * is there a slash after? Yes → route to backend 
  * is there a querystring immediately after? Yes → route to backend
  * is there a hash immediately after? Yes → route to backend
  * is this a beginning of a different and valid pathname starting with the same prefix? Yes → route elsewhere
  
All of this to work around supertokens weird quirk with URLs. 

So in fact I'd say that having a consistent and expected url schema in fact does matter.

rishabhpoddar:
I see. That makes more sense now. Thanks for the explanation. We will sort this out.

rishabhpoddar:
This has been fixed in version >= 0.28.1

PR #627, opened 2022-12-13
fix: add trailing slash to url

alisher-aituarov:
## Summary of change

Trailing slash is added to URL when going back to "Sign In" from "Forgot Password" for consistency

## Related issues

-  https://github.com/supertokens/supertokens-auth-react/issues/625

## Test Plan

## Documentation changes

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`


lib/ts/superTokens.tsx
```diff
@@ -315,6 +316,7 @@ export default class SuperTokens {
             action: "TO_AUTH",
             showSignIn: options.show === "signin",
         });
+        redirectUrl = appendTrailingSlashToURL(redirectUrl);
```
rishabhpoddar, line 319: Instead of changing here, i would suggest that you see where in the SDK we are actually doing a redirection using history or window APIs and modify the URL there to normalise it to add a trailing slash.

This would ensure that any redirection we do (even in the future) will be consistent.

lib/ts/superTokens.tsx
```diff
@@ -315,6 +316,7 @@ export default class SuperTokens {
...
         redirectUrl = appendQueryParamsToURL(redirectUrl, queryParams);
         return this.redirectToUrl(redirectUrl, options.history);
```
rishabhpoddar, line 321: add tests for this PR

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [unreleased]
 
+### Changed
+
+-   Trailing slash is added to URL when going back to "Sign In" from "Forgot Password" for consistency
```
rishabhpoddar, line 12: Add the link to the github issue here

rishabhpoddar (changes requested):

lib/ts/superTokens.tsx
porcellus: I've just discussed this with @alisher-aituarov, and I'd prefer it if didn't normalize this if we get the redirection URL from user overrides. It might be fine for the websiteBasePath (as we know it's a basepath), but not for other links. 
E.g., I want to redirect my users to `/dashboard` after a sign-in - overriding that to redirect to `/dashboard/` is not good.

rishabhpoddar: Hmm ok. But we also need to normalise redirection to password reset page, or to email verification page. So those need to be done manually then in those locations.

rishabhpoddar: Also password reset and email verification links generated on the backend SDK will have to be normalised this way as well?

PR #630, opened 2022-12-14
chore: update supertokens-auth-react version for example apps

alisher-aituarov:
## Summary of change

Updates the supertokens-auth-react version for example apps

## Related issues

-   https://github.com/supertokens/supertokens-core/issues/540

## Test Plan

version updates only

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`


rishabhpoddar:
@alisher-aituarov you need to change the bash branch of the PR to point to the latest X.Y version branch

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [unreleased]
 
+### Changes
```
porcellus, line 10: I'm not sure if this is necessary, but it should be under 0.28 anyway.

porcellus (changes requested):
There is a conflict with the base branch

PR #637, opened 2022-12-29
docs: update with-localstorage example to use header-based auth

porcellus:
## Summary of change

Updates the with-localstorage example to use header-based session from our SDK instead of a custom implementation/

## Related issues

-  https://github.com/supertokens/supertokens-website/pull/173
- https://github.com/supertokens/supertokens-node/pull/440

## Test Plan

N/A

## Documentation changes

https://github.com/supertokens/docs/pull/556

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`


examples/with-localstorage/README.md
```diff
@@ -20,28 +20,6 @@ Storing tokens in localstorage comes with a few drawbacks:
 
 For a detailed comparison please see [our blog](https://supertokens.com/blog/cookies-vs-localstorage-for-sessions-everything-you-need-to-know)
```
rishabhpoddar, line 22: i think this example app can be removed entirely and instead the readme can just point to the docs about token transfer method section.

I know this is localstorage and we store in fronternd cookies, but the point of this example app was to show how users could use with header based auth instead of cookie based.

rishabhpoddar (changes requested):

rishabhpoddar (dismissed):

PR #650, opened 2023-02-01
fix: social buttons style

alisher-aituarov:
## Summary of change

Fixes buttons hover cursor not being changed to pointer issue

## Test Plan

## Documentation changes

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`


porcellus (dismissed):
LGTM

rishabhpoddar:
@alisher-aituarov please resolve conflict

lib/ts/recipe/thirdparty/components/themes/styles.css
```diff
@@ -9,6 +9,10 @@
     padding-bottom: 9px;
 }
 
+[data-supertokens~="providerContainer"] button {
+    cursor: pointer;
```
rishabhpoddar, line 13: can you confirm that if the user provides their own custom component (with a custom provider, or with the google provider), the pointer style is still applied?

rishabhpoddar (changes requested):

lib/ts/styles/styles.css
```diff
@@ -234,6 +234,7 @@
     background-position: center;
     transition: all 0.4s;
```
rishabhpoddar, line 235: is it possible to add unit test for this?

rishabhpoddar (changes requested):

lib/ts/recipe/thirdparty/components/themes/styles.css
alisher-aituarov: it will be applied if user uses button element

rishabhpoddar: The user is not supposed to use the button element in their custom component cause the click handler is applied by us. Please make sure you understand how things work

alisher-aituarov:
Screen Shot 2023-02-02 at 17 00 23
Screen Shot 2023-02-02 at 17 02 17
Screen Shot 2023-02-02 at 17 02 53

lib/ts/styles/styles.css
alisher-aituarov: it is possible, discussed this with @porcellus and this seem to be a lot of work to just to check that line of css does work

porcellus: The other big issue with the test is that we'd be testing that CSS rules work, but as far as I can tell, there is no way to test something like "what's the cursor like if I hover over this part of the page/button".

porcellus (approved):

rishabhpoddar (approved):

PR #653, opened 2023-02-06
fix: keep query redirectToPath's query params when redirecting from auth

alisher-aituarov:
## Summary of change

Makes redirectToPath's query params stay when redirecting to it from auth route

## Related issues

- https://github.com/supertokens/supertokens-web-js/issues/82

## Test Plan

## Documentation changes

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`


porcellus (changes requested):
The changes look good, but please add tests. There are a few `redirectToPath` tests already there that you can use as a base.

test/end-to-end/signin.test.js
```diff
@@ -486,13 +486,15 @@ describe("SuperTokens SignIn", function () {
                 `${TEST_CLIENT_BASE_URL}/redirect-here`
             );
 
-            // test that if we visit auth again, we end up in redirect-heree again
+            // test that if we visit auth again, we end up in redirect-heree again with query params kept
             await Promise.all([
-                page.goto(`${TEST_CLIENT_BASE_URL}/auth?rid=emailpassword&redirectToPath=%2Fredirect-heree`),
+                page.goto(
```
porcellus, line 491: Please add a similar ones that test the redirection:
- emailpassword sign-in 
- emailpassword sign-up.
- passwordless sign-in
- thirdparty sign-in

This test only cheks that the auth pages redirect to the appropriate path if we navigate to it while already logged in.

lib/ts/utils.ts
```diff
@@ -62,14 +62,19 @@ export function getURLHash(): string {
...
         return undefined;
     } else {
         // Prevent Open redirects by normalising path.
         try {
-            return new NormalisedURLPath(param).getAsStringDangerous();
+            const normalisedURLPath = new NormalisedURLPath(param).getAsStringDangerous();
+            if (includeQueryParams === false) {
+                return normalisedURLPath;
+            }
+            const pathQueryParams = param.split("?")[1] ? `?${param.split("?")[1]}` : "";
```
porcellus, line 76: please add a strict undefined check, so we can preserve paths ending in a `?`

porcellus (changes requested):

lib/ts/utils.ts
porcellus: This didn't happen.

lib/ts/utils.ts
```diff
@@ -62,14 +62,19 @@ export function getURLHash(): string {
     return WindowHandlerReference.getReferenceOrThrow().windowHandler.location.getHash().substr(1);
 }
 
-export function getRedirectToPathFromURL(): string | undefined {
+export function getRedirectToPathFromURL({ includeQueryParams } = { includeQueryParams: false }): string | undefined {
```
porcellus, line 65: Is this used anywhere without this param set to true?

porcellus (changes requested):

porcellus (changes requested):
Please remove the parameter of `getRedirectToPathFromURL` if it's always used with the same parameter.

rishabhpoddar (approved):

PR #655, opened 2023-02-13
feat:splitting prebuild UI components

alisher-aituarov:
## Summary of change

Splitting pre built ui components, separating from recipes 

## Test Plan

## Documentation changes

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

-   [ ] Item1
-   [ ] Item2


lib/ts/recipe/recipeRoutes/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+import { RecipeFeatureComponentMap } from "../../types";
+
+export abstract class RecipeRoutes {
+    private pathsToFeatureComponentWithRecipeIdMap?: BaseFeatureComponentMap;
+
+    private static reactRouterDom?: { router: { Route: any }; useHistoryCustom: () => any };
+    private static reactRouterDomIsV6: boolean | undefined = undefined;
+
+    static getRoutes(reactRouterDom: any, instance?: RecipeRoutes): JSX.Element[] {
+        if (instance === undefined) {
```
porcellus, line 17: Since we are throwing if the instance param is `undefined`, I'd prefer if it was a required param and we didn't need this check.

lib/ts/recipe/recipeRoutes/index.tsx
```diff
@@ -0,0 +1,130 @@
+import React from "react";
+import NormalisedURLPath from "supertokens-web-js/lib/build/normalisedURLPath";
+import { RoutingComponent } from "../../components/routingComponent";
+import { getSuperTokensRoutesForReactRouterDom } from "../../components/superTokensRoute";
+import { getSuperTokensRoutesForReactRouterDomV6 } from "../../components/superTokensRouteV6";
+import { BaseFeatureComponentMap, ComponentWithRecipeAndMatchingMethod } from "../../types";
+import { getCurrentNormalisedUrlPath } from "../../utils";
+import { RecipeFeatureComponentMap } from "../../types";
+
+export abstract class RecipeRoutes {
```
porcellus, line 10: Think about renaming this to `RecipeRouter`

lib/ts/recipe/recipeRoutes/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+                "Instance should be passed"
+            );
+        }
+        if (reactRouterDom === undefined) {
+            throw new Error(
+                // eslint-disable-next-line @typescript-eslint/quotes
+                'Please use getSuperTokensRoutesForReactRouterDom like getSuperTokensRoutesForReactRouterDom(require("react-router-dom")) in your render function'
+            );
+        }
+        RecipeRoutes.reactRouterDom = reactRouterDom;
```
porcellus, line 29: We could also check for equality if `RecipeRoutes.reactRouterDom` already has a value. We didn't need this previously because in almost all cases this would only be called once, but this will change now.

porcellus (changes requested):
Very good in general, some smallish refactoring is needed to tighten up the interface.
Plus, please check:
- [ ] What happens if multiple auth recipes are enabled (e.g.: Passwordless and ThirdPartyEmailPassword) and the user navigates to `/auth`
- [ ] What happens if multiple auth recipes are enabled (e.g.: Passwordless and ThirdPartyEmailPassword) and the user navigates to `/auth?rid=...`

alisher-aituarov:
> 

1 - when multiple recipes enabled UI for the recipe that calls getRoutes in JSX will be rendered
2 - navigating to ?rid=recipeId changes nothing

examples/with-emailpassword/src/App.js
```diff
@@ -46,7 +47,7 @@ function App() {
                     <div className="fill">
                         <Routes>
                             {/* This shows the login UI on "/auth" route */}
-                            {getSuperTokensRoutesForReactRouterDom(require("react-router-dom"))}
+                            {EmailPasswordPreBuiltUIRoutes.getRoutes(require("react-router-dom"))}
```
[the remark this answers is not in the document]
alisher-aituarov: do you mean EmailPasswordPreBuiltUIRoutes -> EmailPasswordPreBuiltUI ?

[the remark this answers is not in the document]
alisher-aituarov: renamed to EmailPasswordPreBuiltUI

lib/ts/recipe/emailpassword/preBuiltUI.tsx
```diff
@@ -0,0 +1,126 @@
...
+import SignInAndUpTheme from "./components/themes/signInAndUp";
+import ResetPasswordUsingTokenTheme from "./components/themes/resetPasswordUsingToken";
+
+export class EmailPasswordPreBuiltUIRoutes extends RecipeRoutes {
+    private constructor(private readonly recipeInstance: EmailPassword) {
+        super();
+    }
+    static instance: EmailPasswordPreBuiltUIRoutes;
+
+    static init(recipeInstance: EmailPassword): void {
```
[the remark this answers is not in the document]
porcellus: 2. Yes, but we only need them internally. This class is exposed to the user, we do not want them to be able to call `getInstanceOrInitAndGetInstance` or those two functions.

porcellus (changes requested):
I'll think about this, but the behavior you described doesn't match the current one.

On a separate note, the merge commit is showing up for some reason in the diff of the PR review, making it very hard to review more broadly. I think this could be fixed by a rebase, or merging the target branch again, I'm not sure. Please try and check it out.

lib/ts/recipe/recipeRoutes/index.tsx
porcellus: This hasn't happened

porcellus (changes requested):
We need to replicate the old way multiple auth recipes used to work: check out the `with-thirdpartyemailpassword-passwordless` example.

##### Requirements:

If the user visits a path and has an `rid` query param:
- the prebuilt UI with the matching `rid` should be shown

If the user visits a path and has no `rid` query param:
- the first prebuilt UI that can handle the path should be shown

Below is a solution, but it's not perfect, so see if you can come up with anything better/simpler:
- It might add a delay in rendering (you could try and test this) because we need to update the underlying context before rendering the current UI
- Ideally we'd also detect if there is no recipe/prebuilt-UI added with the `rid` in the query params and just handle it as if it wasn't there.
- This requires all apps to be wrapped by `SuperTokensWrapper` (not a big issue)

##### Solution:

I think this is doable by adding a new type of context into `SuperTokensWrapper` that stores the `rid` of the recipe currently handling the request and the path. 

This way, the prebuilt UIs can:
1. Check if the current URL has an `rid` defined - if it does, check if it matches their recipe and render the appropriate UI.
2. Check if the context contains the current URL and the `rid` of their recipe. If it does, they should render the UI.
3. If the current URL has no `rid` defined, they should try to add their `rid` along the current URL into the context if it's not defined or contains a different URL.

test/unit/recipe/emailpassword/emailPassword.test.tsx
```diff
@@ -97,8 +99,9 @@ describe("EmailPassword", function () {
                 disableDefaultUI: true,
             },
         })(SuperTokens.getInstanceOrThrow().appInfo, false);
-        assert(EmailPassword.getInstanceOrThrow().getFeatures()["/auth"] === undefined);
-        assert(EmailPassword.getInstanceOrThrow().getFeatures()["/auth/reset-password"] === undefined);
+        // console.log(EmailPassword.getInstanceOrThrow().config);
```
porcellus, line 102: This can be removed.

porcellus:
Please add the preBuiltUI entries into size limits

CHANGELOG.md
```diff
@@ -7,6 +7,120 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+    ],
+});
+
+<SuperTokensWrapper>
+    <div className="App">
+        <Router>
+            <div className="fill">
+                <Routes>
+                    {/* This shows the login UI on "/auth" route for Passwordless recipe */}
+                    {PasswordlessPreBuiltUI.getReactRouterDomRoutes(require("react-router-dom"))}
```
porcellus, line 78: These needs to be updated

porcellus (changes requested):
- examples also need to be updated

lib/ts/recipe/recipeRouter/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+            }
+
+            pathsToFeatureComponentWithRecipeIdMap[featurePath].push(features[featurePath]);
+        }
+
+        this.pathsToFeatureComponentWithRecipeIdMap = pathsToFeatureComponentWithRecipeIdMap;
+        return this.pathsToFeatureComponentWithRecipeIdMap;
+    };
+
+    getMatchingComponentForRouteAndRecipeId = (
```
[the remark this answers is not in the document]
porcellus: Please remove it if it's not required.

examples/with-next-iframe/pages/auth/[[...path]].js
```diff
@@ -3,13 +3,20 @@ import React, { useEffect } from "react";
 import styles from "../../styles/Home.module.css";
 import dynamic from "next/dynamic";
 import SuperTokens from "supertokens-auth-react";
+import { EmailVerificationPreBuiltUI } from "supertokens-auth-react/recipe/emailverification/prebuiltui";
+import { EmailPasswordPreBuiltUI } from "supertokens-auth-react/recipe/emailpassword/prebuiltui";
 import { redirectToAuth } from "supertokens-auth-react";
 
-const SuperTokensComponentNoSSR = dynamic(new Promise((res) => res(SuperTokens.getRoutingComponent)), { ssr: false });
+const SuperTokensComponentNoSSR = dynamic(
+    new Promise((res) => res(() => getRoutingComponent([EmailVerificationPreBuiltUI, EmailPasswordPreBuiltUI]))),
```
Comment (porcellus) [inline examples/with-next-iframe/pages/auth/[[...path]].js:11]: I don't think this was tested, `getRoutingComponent` is not imported.

CHANGELOG.md
```diff
@@ -7,6 +7,121 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+```tsx
+import { SuperTokensWrapper } from "supertokens-auth-react";
+import { PasswordlessPreBuiltUI } from "supertokens-auth-react/recipe/passwordless/prebuiltui";
+
+class App extends React.Component {
+    render() {
+        if (PasswordlessPreBuiltUI.canHandleRoute()) {
+            // This renders the login UI on the /auth route
+            return PasswordlessPreBuiltUI.getRoutingComponent();
```
porcellus, line 117: This needs to be updated as well.

porcellus (changes requested):

porcellus (changes requested):

porcellus:
The changes look good; Please merge this into the combined branch, then close this PR.

PR #696, opened 2023-05-26
feat: Add basic support for rendering pre built UI components in SSR environments

nkshah2:
## Summary of change

The current problem of using auth react in SSR frameworks such as NextJS is that the init call requires the window object which is not present on the server side. Using useEffect solves that but results in the server side render failing because none of the recipes have been initialised so wrapper components exposed by the SDK throw errors.

This PR tries to solve the problem of throwing an error on the server side by returning an empty div instead of throwing if the recipe has not been initialised and the window object is not present (for pre built UI components).

- Adds a wrapper component to render an empty div if there is an error when getting the recipe instance and window is not present
- Adds a util function to return an empty map for the feature map if the recipe has not been initialised and window is not present
- Modifies email verification pre built UI to use the wrapper component and util function
- Modified email verification pre built UI to use the static `getFeatureComponent` variant instead of the instance one when exposing the `EmailVerification` to the user
- Modifies the index file from the ts/ui folder to use similar logic (Is this required if all recipes internally handle this?)

Note: During development rollup kept throwing errors for older types that had been removed. This PR also modifies the rollup config to disable cache

## Related issues
- ...

## Test Plan
WIP

## Documentation changes
NA

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

-   [ ] ...


lib/ts/ui/index.tsx
```diff
@@ -85,7 +85,20 @@ export default class UI {
     }
 
     static getRoutingComponent(preBuiltUiClassList: PreBuiltRecipes): JSX.Element {
-        const recipeList = preBuiltUiClassList.map((r) => r.getInstanceOrInitAndGetInstance());
+        let recipeList: RecipeRouter[];
+
+        try {
+            recipeList = preBuiltUiClassList.map((r) => r.getInstanceOrInitAndGetInstance());
+        } catch (e) {
+            // TODO during review: Is this checking required here if all recipes internally handle it?
```
porcellus, line 93: If `RoutingComponent` can handle this (and I think it should be fine) then there is no need to do this here.

lib/ts/ui/uiutils.tsx
```diff
@@ -0,0 +1,22 @@
...
+    try {
+        recipeInstance = getRecipeInstance();
+    } catch (e) {
+        // eslint-disable-next-line supertokens-auth-react/no-direct-window-object
+        if (typeof window === "undefined") {
+            return {};
+        }
+
+        throw e;
+    }
```
porcellus, line 19: While this solves the immediate issue, I think I'm not sure how much this renders in SSR (if at all).
- This change means that we are adding different routes (and components) during SSR and browser rendering. This could cause hydration issues
- Please check that at least some SSR happens (e.g.: by checking the preview of the html loaded by during the initial page load of `/auth`)

porcellus (changes requested):

nkshah2:
Closing this because we went for a different approach

PR #715, opened 2023-06-12
refactor: Allow SDK to be initialised without window being defined

nkshah2:
## Summary of change

- Refactors the init logic to not throw if window is not defined

## Related issues

- ...

## Documentation changes

WIP

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

-   [ ] Update CHANGELOG and package version
- [ ] Update install path for web-js once its released

### DIFF
--- a/lib/ts/utils.ts
+++ b/lib/ts/utils.ts
@@ -180,7 +180,7 @@ export async function validateForm(
  */
 export function getCurrentNormalisedUrlPath(): NormalisedURLPath {
     try {
-        WindowHandlerReference.getReferenceOrThrow();
+        WindowHandlerReference.getReferenceOrThrow().windowHandler.getWindowUnsafe();
     } catch (_) {
         return new NormalisedURLPath("/");
     }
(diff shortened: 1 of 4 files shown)


lib/ts/utils.ts
```diff
@@ -179,6 +179,12 @@ export async function validateForm(
  * getCurrentNormalisedUrlPath
  */
 export function getCurrentNormalisedUrlPath(): NormalisedURLPath {
+    try {
+        WindowHandlerReference.getReferenceOrThrow().windowHandler.getWindowUnsafe();
+    } catch (_) {
+        return new NormalisedURLPath("/");
```
porcellus, line 185: I don't think this is correct. For example, this'll make `canHandleRoute` return false everywhere if the `basePath` is not `/` and true everywhere if it is. Not sure how bad this is but it can cause some unexpected things.
This is only used in a few places, where we could make it overrideable by adding an optional param that'll skip calling this.

porcellus (changes requested):

PR #725, opened 2023-07-26
feat: Delaying the initial `getDynamicLoginMethods`

porcellus:
## Summary of change

-   Delaying the initial `getDynamicLoginMethods` until the first feature/routing component renders
-   Added a spinner that shows up if `usesDynamicLoginMethods` is true and `getDynamicLoginMethods` takes longer than 200ms
-   Updated with-one-login-many-subdomains example to match the create-supertokens-app-example

## Related issues

-   

## Test Plan

- Updated existing tests
- Added a test case to check the spinner shows up

## Documentation changes

Done separately

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`


CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [unreleased]
```
rishabhpoddar, line 8: - size check failed
- Is this change fully backwards compatible? If so, please update patch version, else we can discuss

rishabhpoddar (changes requested):

Issue #36, opened 2020-11-13
Field decorators


kant01ne:
## Field decorators

We want to allow to extend the default EmailPassword theme to add custom behaviour such as :
 -  Password strength meters
     - weak | good | excellent indicators
     - password time to crack
     - Password is in top 100/1k most used password
 - Email suggestions: "We noticed that you are using a Gmail account, do you have a work email?"

We want to do this without having to:
  - Implement a new theme that behaves exactly the same as the default except for that one small behaviour.
  - Ask customers to disable default implementation to overwrite the default theme.

### Solution

The idea is to add an optional property called `decorators` to `formFields`. This decorator consists of a React element that will be inserted below the input field and that will take its value as props.

Example:

```js
SuperTokens.init({
  appInfo: {},
  recipeList: [
    EmailPassword.init({
      signInAndUpFeature: {
        onSuccessRedirectURL: '/dashboard',
        signUpForm: {
          formFields: [{
              id: "password",
              decorator: <PasswordStrengthMeterDecorator />
            }]
         }
       }
    })
   ] 
});
```

With `PasswordStrengthMeterDecorator` a React element that takes a value as props (the pasword which is updated on each changes), and displays a UI below the input.
We can provide few decorators as examples but the idea is that the developers can implement their own to extend the default theme easily if they like.

### Implementation

Implementation is straightforward, we simply need to:
  - Add `decorator?: JSX.Element` to `FormFieldBaseConfig` type.
  - Prevent propagating SignUp decorators to SignIn
  - Add `onChange in `input.tsx` to make sure that the state is updated each time the value is updated (as opposed to using React refs).
  - Insert the `decorator` element, give it input `value` as props.

rishabhpoddar:
This is a good idea. However, I think we should postpone this to post the MVP launch.  
- If someone adds a decorator, will the formField error message still show in case of errors? If yes, how will the decorator be placed / behave in that case?
- Shouldn't the ability to add decorater go as a prop to the theme. I feel it's better there since having it in init forces all themes to take care of this, which may not be appropriate.

The above being said, I think the problem of wanting to change something very slightly resulting in having to build the whole theme yourself is deeper issue that may require a different solution. 

Happy to discuss this after the MVP release.

NkxxkN:
> If someone adds a decorator, will the formField error message still show in case of errors? If yes, how will the decorator be placed / behave in that case?

Decorators are interactive. They update every time their respective value changes. They are inserted before the error field.

[...]

PR #756, opened 2023-11-06
feat: add getDefaultValue to signin & signup form fields  

amitbadala:
## Summary of change

This PR enables getDefaultValue prop on signing form fields for `emailpassword` recipe
Users can now set default values for both sign in & sign up form fields 

## Related issues

-   [Field decorators #36](https://github.com/supertokens/supertokens-auth-react/issues/36)

## Test Plan

## Documentation changes

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

### DIFF
--- a/examples/for-tests/src/App.js
+++ b/examples/for-tests/src/App.js
@@ -740,6 +740,17 @@ function getSignInFormFields() {
     return {};
 }
 
+function emailFormForResetPassword() {
+    let showDefaultFields = localStorage.getItem("SHOW_DEFAULT_EMAIL_FOR_RESET_PASSWORD");
+    if (showDefaultFields === "YES") {
+        return {
+            getDefaultValue: () => "company@email.com",
+            style: theme,
+        };
+    }
+    return { style: theme };
+}
+
 function getEmailPasswordConfigs({ disableDefaultUI }) {
     return EmailPassword.init({
         style: `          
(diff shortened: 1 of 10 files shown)


examples/for-tests/src/App.js
```diff
@@ -721,6 +721,25 @@ function getFormFields() {
     return formFields;
 }
 
+function getSignInFormFields() {
```
rishabhpoddar, line 724: we also need to test the reset password screen can have a default value in a similar way

rishabhpoddar (changes requested):

lib/ts/recipe/emailpassword/types.ts
```diff
@@ -198,7 +198,9 @@ export type ResetPasswordUsingTokenUserInput = {
     /*
      * enterEmailForm config.
      */
-    enterEmailForm?: FeatureBaseConfig;
+    enterEmailForm?: FeatureBaseConfig & {
+        getDefaultValue?: () => string;
```
rishabhpoddar, line 202: no.. this does not go here. It should already be a part of the formField config. Why have you added this here?

rishabhpoddar (changes requested):

lib/ts/recipe/emailpassword/types.ts
amitbadala: this is for user input 
formField was never part of the emailForm Input
It gets normalized into formField type later for resetPassword

PR #757, opened 2023-11-06
feat: add nonOptionalErrorMsg props 

amitbadala:
## Summary of change

Modified the sign-in and sign-up form fields to accept a new property: `nonOptionalErrorMsg`. 
This allows for the display of a custom error message specified by the user, instead of the default `Field is not Optional` message, for fields that are mandatory.

## Related issues

-   [Field decorators #36](https://github.com/supertokens/supertokens-auth-react/issues/36)

## Test Plan

## Documentation changes

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

- [ ] Check in all backend SDKs that "Field is not optional" is the right message that's returned from the backend.

### DIFF
--- a/lib/ts/recipe/emailpassword/components/library/formBase.tsx
+++ b/lib/ts/recipe/emailpassword/components/library/formBase.tsx
@@ -251,10 +251,21 @@ export const FormBase: React.FC<FormBaseProps<any>> = (props) => {
                     // If field error.
                     if (result.status === "FIELD_ERROR") {
                         const errorFields = result.formFields;
-
-                        setFieldStates((os) =>
-                            os.map((fs) => ({ ...fs, error: errorFields.find((ef: any) => ef.id === fs.id)?.error }))
-                        );
+                        const getErrorMessage = (fs: FieldState) => {
+                            const errorMessage = errorFields.find((ef: any) => ef.id === fs.id)?.error;
+                            if (errorMessage === "Field is not optional") {
+                                const fieldConfigData = props.formFields.find((f) => f.id === fs.id);
+                                // replace non-optional server error message from nonOptionalErrorMsg
+                                if (
+                                    fieldConfigData?.nonOptionalErrorMsg !== undefined &&
+                                    fieldConfigData?.nonOptionalErrorMsg !== ""
+                                ) {
+                                    return fieldConfigData?.nonOptionalErrorMsg;
+                                }
+                            }
+                            return errorMessage;
+                        };
+                        setFieldStates((os) => os.map((fs) => ({ ...fs, error: getErrorMessage(fs) })));
                     }
                 }
             } catch (e) {
--- a/lib/ts/recipe/emailpassword/components/library/formBase.tsx
+++ b/lib/ts/recipe/emailpassword/components/library/formBase.tsx
@@ -256,10 +256,7 @@ export const FormBase: React.FC<FormBaseProps<any>> = (props) => {
                             if (errorMessage === "Field is not optional") {
                                 const fieldConfigData = props.formFields.find((f) => f.id === fs.id);
                                 // replace non-optional server error message from nonOptionalErrorMsg
-                                if (
-                                    fieldConfigData?.nonOptionalErrorMsg !== undefined &&
-                                    fieldConfigData?.nonOptionalErrorMsg !== ""
-                                ) {
+                                if (fieldConfigData?.nonOptionalErrorMsg !== undefined) {
                                     return fieldConfigData?.nonOptionalErrorMsg;
                                 }
                             }
--- a/lib/ts/recipe/emailpassword/utils.ts
+++ b/lib/ts/recipe/emailpassword/utils.ts
@@ -340,7 +340,7 @@ export function mergeFormFields(
 export function getFormattedFormField(field: NormalisedFormField): NormalisedFormField {
     // Fields with the 'nonOptionalErrorMsg' property must have a valid message defined
     if (field.optional === false && field.nonOptionalErrorMsg === "") {
-        throw new Error(`nonOptionalErrorMsg for field ${field.id} cannot be empty`);
+        throw new Error(`nonOptionalErrorMsg for field ${field.id} cannot be an empty string`);
     }
     return {
         ...field,
--- a/lib/ts/recipe/emailpassword/utils.ts
+++ b/lib/ts/recipe/emailpassword/utils.ts
@@ -326,10 +326,10 @@ export function mergeFormFields(
         // If new field, push to mergeFormFields.
         if (isNewField) {
             mergedFormFields.push({
-                optional: false,
                 placeholder: userField.label,
                 validate: defaultValidate,
                 ...userField,
+                optional: Boolean(userField.optional),
             });
         }
     }
--- a/lib/ts/recipe/emailpassword/utils.ts
+++ b/lib/ts/recipe/emailpassword/utils.ts
@@ -343,6 +343,9 @@ export function getFormattedFormField(field: NormalisedFormField): NormalisedFor
         validate: async (value: any): Promise<string | undefined> => {
             // Absent or not optional empty field
             if (value === "" && field.optional === false) {
+                if (field.nonOptionalErrorMsg !== undefined && field.nonOptionalErrorMsg !== "") {
+                    return field.nonOptionalErrorMsg;
+                }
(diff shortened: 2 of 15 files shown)


lib/ts/recipe/emailpassword/components/library/formBase.tsx
```diff
@@ -251,10 +251,21 @@ export const FormBase: React.FC<FormBaseProps<any>> = (props) => {
...
-                        );
+                        const getErrorMessage = (fs: FieldState) => {
+                            const errorMessage = errorFields.find((ef: any) => ef.id === fs.id)?.error;
+                            if (errorMessage === "Field is not optional") {
+                                const fieldConfigData = props.formFields.find((f) => f.id === fs.id);
+                                // replace non-optional server error message from nonOptionalErrorMsg
+                                if (
+                                    fieldConfigData?.nonOptionalErrorMsg !== undefined &&
+                                    fieldConfigData?.nonOptionalErrorMsg !== ""
+                                ) {
```
rishabhpoddar, line 262: Why are you checking for `!== ""`?

rishabhpoddar: What happens if the user sets it to `""`? Does it not render the optional error at all?

lib/ts/recipe/emailpassword/components/library/formBase.tsx
```diff
@@ -251,10 +251,21 @@ export const FormBase: React.FC<FormBaseProps<any>> = (props) => {
                     // If field error.
                     if (result.status === "FIELD_ERROR") {
                         const errorFields = result.formFields;
-
-                        setFieldStates((os) =>
-                            os.map((fs) => ({ ...fs, error: errorFields.find((ef: any) => ef.id === fs.id)?.error }))
-                        );
+                        const getErrorMessage = (fs: FieldState) => {
+                            const errorMessage = errorFields.find((ef: any) => ef.id === fs.id)?.error;
+                            if (errorMessage === "Field is not optional") {
```
rishabhpoddar, line 256: check in all backend SDKs that "Field is not optional" is the right message that's returned from the backend.

lib/ts/recipe/emailpassword/utils.ts
```diff
@@ -343,6 +343,9 @@ export function getFormattedFormField(field: NormalisedFormField): NormalisedFor
         validate: async (value: any): Promise<string | undefined> => {
             // Absent or not optional empty field
             if (value === "" && field.optional === false) {
+                if (field.nonOptionalErrorMsg !== undefined && field.nonOptionalErrorMsg !== "") {
```
rishabhpoddar, line 346: why are you checking for `!== ""`?

test/end-to-end/signin.test.js
```diff
@@ -757,6 +758,25 @@ describe("SuperTokens SignIn", function () {
...
+                window.localStorage.setItem("SHOW_SIGNIN_WITH_NON_OPTIONAL_ERROR_MESSAGE", "YES")
+            );
+            await page.reload({
+                waitUntil: "domcontentloaded",
+            });
+
+            await submitForm(page);
+            let formFieldErrors = await getFieldErrors(page);
+            // Also standard non-optional-error is displayed if nonOptionalErrorMsg is not provided
+            assert.deepStrictEqual(formFieldErrors, ["Please add email", "Field is not optional"]);
```
rishabhpoddar, line 777: you should also enforce that there is no network call being made, and that this error is due to frontend checking.

rishabhpoddar (changes requested):

test/end-to-end/signup.test.js
```diff
@@ -476,6 +476,75 @@ describe("SuperTokens SignUp", function () {
...
+
+        it("Check on blank form submit nonOptionalErrorMsg gets displayed as expected", async function () {
+            await submitForm(page);
+            let formFieldErrors = await getFieldErrors(page);
+            // Also standard non-optional-error is displayed if nonOptionalErrorMsg is not provided
+            assert.deepStrictEqual(formFieldErrors, [
+                "Field is not optional",
+                "Field is not optional",
+                "You must accept the terms and conditions",
+            ]);
```
rishabhpoddar, line 488: you should also enforce that there is no network call being made, and that this error is due to frontend checking.

rishabhpoddar (changes requested):

lib/ts/recipe/emailpassword/components/library/formBase.tsx
amitbadala: the user will get the standard usual error  `Field is not optional` 
if the value of nonOptionalErrorMsg is set to empty string

lib/ts/recipe/emailpassword/utils.ts
amitbadala: the user will get the standard usual error Field is not optional
if the value of nonOptionalErrorMsg is set to empty string

lib/ts/recipe/emailpassword/components/library/formBase.tsx
rishabhpoddar: Yea, but why are we doing that?

amitbadala: if we don't handle that 
- The end-user won't be able to submit the form & won't realize because of empty string

If we handle it
- It conveys the missing data in a usual way.

lib/ts/recipe/emailpassword/utils.ts
```diff
@@ -338,12 +338,16 @@ export function mergeFormFields(
 }
 
 export function getFormattedFormField(field: NormalisedFormField): NormalisedFormField {
+    // Fields with the 'nonOptionalErrorMsg' property must have a valid message defined
+    if (field.optional === false && field.nonOptionalErrorMsg === "") {
```
rishabhpoddar, line 342: what if `optional` is `undefined`? Does that count as being `false` or `true`? Or will `flag.optional` never be `undefined` here?

lib/ts/recipe/emailpassword/components/library/formBase.tsx
```diff
@@ -251,10 +251,21 @@ export const FormBase: React.FC<FormBaseProps<any>> = (props) => {
...
+                                const fieldConfigData = props.formFields.find((f) => f.id === fs.id);
+                                // replace non-optional server error message from nonOptionalErrorMsg
+                                if (
+                                    fieldConfigData?.nonOptionalErrorMsg !== undefined &&
+                                    fieldConfigData?.nonOptionalErrorMsg !== ""
```
rishabhpoddar, line 261: you are still checking for `!== ""` why?

lib/ts/recipe/emailpassword/utils.ts
```diff
@@ -338,11 +338,18 @@ export function mergeFormFields(
...
+        throw new Error(`nonOptionalErrorMsg for field ${field.id} cannot be empty`);
```
rishabhpoddar, line 343:
```suggestion
        throw new Error(`nonOptionalErrorMsg for field ${field.id} cannot be an empty string`);
```

rishabhpoddar (changes requested):

lib/ts/recipe/emailpassword/utils.ts
amitbadala: `flag.optional `will be `undefined` only if the user explicitly sets it
by default it will be `false`

it doesn't count as `false`

rishabhpoddar: I think undefined means that it is false.. so you have to check `if ((field.optional === false || field.optional === undefined) && field.nonOptionalErrorMsg === "")) {` right?

lib/ts/recipe/emailpassword/utils.ts
```diff
@@ -326,10 +326,10 @@ export function mergeFormFields(
         // If new field, push to mergeFormFields.
         if (isNewField) {
             mergedFormFields.push({
-                optional: false,
                 placeholder: userField.label,
                 validate: defaultValidate,
                 ...userField,
+                optional: Boolean(userField.optional),
```
rishabhpoddar, line 332: this change is not needed.. if you see, it's already being set to false by default.

rishabhpoddar (changes requested):

PR #760, opened 2023-11-11
Custom form fields

rishabhpoddar:
## Summary of change

- Exposes inputComponent type in the form fields input which allows users to add custom react components as form fields for the sign up form. Note that this is only exposed to the sign up form.
- The functions passed to the inputComponent, like onChange, now just take the value. Whereas before, they used to take form id and value. This change is applied to other functions like onInputBlur etc as well.
- Add a getDefaultValue function as an input to each formField as well. This allows users to pre fill the form field value. This is a function because it allows more flexibility, for example if the value must be read from localstorage, or query param based on some condition.
- Adds another form field input called `nonOptionalErrorMsg`, which allows users to replace the default "Field is not optional" error with something else, on a per form field basis. For example, if a form field is a checkbox, then instead of showing "Field is not optional", we can show "Please check this before signing up"

## Related issues

-   https://github.com/supertokens/supertokens-auth-react/issues/36

## Test Plan

TODO

## Documentation changes

TODO

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`


CHANGELOG.md
```diff
@@ -5,6 +5,46 @@ All notable changes to this project will be documented in this file.
...
+### Added
+
+-   Introduced the capability to utilize custom components in the Email-Password based recipes' signup form fields by exposing inputComponent types.
+-   Implemented the functionality to assign default values to the form fields in the Email-Password based recipes.
+-   Simplified onChange prop usage in inputComponent - id attribute removed.
+
+Following is an example of how to use above features.
+
+```tsx
+EmailPassword.init({
```
rishabhpoddar, line 19: add an example with thirdpartyemailpassword.

lib/ts/recipe/emailpassword/components/library/input.tsx
```diff
@@ -20,7 +20,6 @@ import CheckedIcon from "../../../../components/assets/checkedIcon";
 import ErrorIcon from "../../../../components/assets/errorIcon";
```
rishabhpoddar, line 20: if the label is `""` for a form field, and it's necessary, do not show the `*` char

test/end-to-end/signin.test.js
```diff
@@ -669,6 +670,138 @@ describe("SuperTokens SignIn", function () {
...
+            ];
+
+            let assertionError = null;
+            let interceptionPassed = false;
+
+            const requestHandler = async (request) => {
+                if (request.url().includes(SIGN_IN_API) && request.method() === "POST") {
+                    try {
+                        const postData = JSON.parse(request.postData());
+                        expectedDefautlValues.forEach(({ id, value }) => {
```
porcellus, line 715: I'd prefer a `for..of`, but this is OK.

test/end-to-end/signin.test.js
```diff
@@ -669,6 +670,138 @@ describe("SuperTokens SignIn", function () {
...
+                // Also standard non-optional-error is displayed if nonOptionalErrorMsg is not provided
+                assert.deepStrictEqual(formFieldErrors, ["Please add email", "Field is not optional"]);
+            } finally {
+                page.off("request", requestHandler);
+                await page.setRequestInterception(false);
+            }
+
+            if (apiCallMade) {
+                throw new Error("Empty form making API request to signin");
+            }
```
porcellus, line 802: Please use assert instead.

test/end-to-end/signin.test.js
```diff
@@ -669,6 +670,138 @@ describe("SuperTokens SignIn", function () {
...
+
+            if (!interceptionPassed) {
+                throw new Error("test failed");
+            }
+        });
+    });
+
+    describe("Check if nonOptionalErrorMsg works as expected", function () {
+        it("Check on blank form submit nonOptionalErrorMsg gets displayed as expected", async function () {
```
porcellus, line 763: Not really an issue but these are worded differently from the above.
```suggestion
    describe("nonOptionalErrorMsg", function () {
        it("should be displayed on a blank form submit", async function () {
```

test/helpers.js
```diff
@@ -417,6 +417,17 @@ export async function setInputValues(page, fields) {
     return await new Promise((r) => setTimeout(r, 300));
 }
 
+export async function setSelectDropdownValue(page, selector, optionValue) {
+    return await page.evaluate(
+        ({ selector, ST_ROOT_SELECTOR, optionValue }) => {
+            const select = document.querySelector(ST_ROOT_SELECTOR).shadowRoot.querySelector(selector);
+            select.value = optionValue;
+            select.dispatchEvent(new Event("change", { bubbles: true }));
```
porcellus, line 425: Is there no way of doing this with `page.select`?

test/end-to-end/thirdpartyemailpassword.test.js
```diff
@@ -440,4 +442,498 @@ describe("SuperTokens Third Party Email Password", function () {
...
+
+            const requestHandler = (request) => {
+                const url = request.url();
+                if (url === SIGN_IN_API) {
+                    apiCallMade = true;
+                    request.continue();
+                } else {
+                    request.continue();
+                }
+            };
```
porcellus, line 492: Please check but, I think `on("request", ...` fires without `setRequestInterception(true)`. We only need it to change/delay the responses.

porcellus (changes requested):

lib/ts/recipe/emailpassword/utils.ts
```diff
@@ -338,11 +338,18 @@ export function mergeFormFields(
 }
 
 export function getFormattedFormField(field: NormalisedFormField): NormalisedFormField {
+    // Fields with the 'nonOptionalErrorMsg' property must have a valid message defined
+    if (field.optional === false && field.nonOptionalErrorMsg === "") {
+        throw new Error(`nonOptionalErrorMsg for field ${field.id} cannot be an empty string`);
+    }
```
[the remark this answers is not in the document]
amitbadala: nonOptionalErrorMsg is not required even if the field is not optional
here, we enforce a valid value for nonOptionalErrorMsg if it has been supplied.

lib/ts/version.ts
```diff
@@ -12,4 +12,4 @@
  * License for the specific language governing permissions and limitations
  * under the License.
  */
-export const package_version = "0.35.6";
+export const package_version = "0.36.0";
```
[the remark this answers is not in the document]
amitbadala: Nope it's not a breaking change
will update to `0.35.7`

test/helpers.js
amitbadala: Puppeteer's page.select() method doesn't natively support selection within shadow roots, 
hence using this approch

test/end-to-end/signin.test.js
```diff
@@ -669,6 +670,138 @@ describe("SuperTokens SignIn", function () {
             });
         });
     });
+
+    describe("SignIn default field tests", function () {
```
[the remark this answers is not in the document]
amitbadala: Yup test fails

lib/ts/recipe/emailpassword/utils.ts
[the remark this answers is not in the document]
porcellus: Oh right, it's just an empty string check, it can be undefined. I misread that, sorry.

test/end-to-end/signin.test.js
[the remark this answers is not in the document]
porcellus: Check out this file: `test/end-to-end/thirdpartypasswordless.tp.test.js`
It both shows how you can check if we are running in react 16 and how we skip tests.

test/helpers.js
porcellus: not 100 sure, but please check if something like this works:
```
    const ele = await waitForSTElement(page, "[data-supertokens~=selectX]");
    ele.select(...)
 ```

Issue #761, opened 2023-11-13
Abstract away navigation from the pre built UI


rishabhpoddar:
Allows users to define how page navigation should be done, or if at all across our components. This means that we also allow users to embed the entire signipin experience if they want to in a modal and that should not do any redirection during the process, or after.
Specifically, we want to:
- [ ] Allow returning `null` from getRedirectionURL, which does not do any navigation. Keep in mind that if this is done, for the sign in success event, then our component should not constantly try and redirect (cause it's wrapper around a component that redirects if a session exists)
- [ ] Add prop to return of getRoutingComponent() which accepts the path (instead of using the global path), so users can control which UI to show.
- [ ] Investigate how clerk components work in terms of embedding.
- [ ] Allowing users to use their own navigation lib instead of react router dom

PR #766, opened 2023-11-23
feat: Allow returning null from getRedirectionURL override

anku255:
## Summary of change

This PR updates `getRedirectionURL` signature to allow returning `null` that would prevent redirection. 

## Related issues

-  https://github.com/supertokens/supertokens-auth-react/issues/761

## Test Plan

I have written one unit test and added some code in `with-typescript` repo to test the function signature. Also added e2e tests for emailpassword and passwordless signups.

## Documentation changes

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

-   [x] Write e2e test
-   [x] Update package version
-   [ ] Update documenation
-   [x] Remove success screen
-   [x] Add a setInterval that runs every 2s and redirect if session exists in passwordless
-   [ ] Update [success screen](https://supertokens.com/docs/passwordless/pre-built-ui/further-reading/passwordless-login#logging-in-via-otp-and-magic-link-simultaneously) doc


test/with-typescript/src/App.tsx
```diff
@@ -1460,3 +1460,44 @@ Multitenancy.init({
         }),
     },
 });
+
+// Testing that 'null' is allowed to be returned from getRedirectionURL
+SuperTokens.init({
```
rishabhpoddar, line 1465: also add here that null is allowed to be returned from session claim's on failure redirection as well.

rishabhpoddar: Something like this:

```
import { BooleanClaim } from "supertokens-auth-react/recipe/session";

export const PhoneVerifiedClaim = new BooleanClaim({
    id: "phone-verified",
    refresh: async () => {
        // This is something we have no way of refreshing, so this is a no-op
    },
    onFailureRedirection: () => null,
});
```

test/unit/recipe/emailpassword/emailPassword.test.tsx
```diff
@@ -626,4 +626,18 @@ describe("EmailPassword", function () {
             }
         }
     });
+
```
rishabhpoddar, line 629: Missing e2e tests

rishabhpoddar, line 629: Modfy changelog and version as well - since this is a non breaking change, it can be 0.35.8 (there is already 0.35.7 that is in the process of being released)

rishabhpoddar (changes requested):

test/with-typescript/src/App.tsx
anku255: done

test/unit/recipe/emailpassword/emailPassword.test.tsx
anku255: done

anku255: done

examples/for-tests/src/App.js
```diff
@@ -816,6 +816,9 @@ function getEmailPasswordConfigs({ disableDefaultUI, formFieldType }) {
             console.log(`ST_LOGS EMAIL_PASSWORD GET_REDIRECTION_URL ${context.action}`);
             if (context.action === "SUCCESS") {
                 setIsNewUserToStorage("emailpassword", context.isNewRecipeUser);
+                if (testContext.disableRedirectionAfterSuccessfulSignInUp) {
```
rishabhpoddar, line 819: need to also modify test with react 16

test/end-to-end/getRedirectionURL.test.js
```diff
@@ -200,7 +200,6 @@ describe("getRedirectionURL Tests", function () {
                 if (!_isPasswordlessSupported) {
                     didSkip = true;
                     this.skip();
-                    return;
```
rishabhpoddar, line 203: why is this removed?

test/end-to-end/getRedirectionURL.test.js
```diff
@@ -287,7 +286,6 @@ describe("getRedirectionURL Tests", function () {
                 if (!_isThirdPartyPasswordlessSupported) {
                     didSkip = true;
                     this.skip();
-                    return;
```
rishabhpoddar, line 290: why is this removed?

test/end-to-end/getRedirectionURL.test.js
```diff
@@ -377,5 +375,152 @@ describe("getRedirectionURL Tests", function () {
...
+                    browser = await puppeteer.launch({
+                        args: ["--no-sandbox", "--disable-setuid-sandbox"],
+                        headless: true,
+                    });
+
+                    page = await browser.newPage();
+                    // We need to set the localStorage value before the page loads to ensure ST initialises with the correct value
+                    await page.evaluateOnNewDocument(() => {
+                        localStorage.setItem("disableRedirectionAfterSuccessfulSignInUp", "true");
+                        localStorage.removeItem("isNewUserCheck");
```
rishabhpoddar, line 400: why is this required? What does the `isNewUserCheck` do?

test/end-to-end/getRedirectionURL.test.js
```diff
@@ -377,5 +375,152 @@ describe("getRedirectionURL Tests", function () {
...
+                    const urlBeforeSignUp = await page.url();
+                    await defaultSignUp(page);
+                    const urlAfterSignUp = await page.url();
+
+                    const newUserCheck = await page.evaluate(() => localStorage.getItem("isNewUserCheck"));
+                    assert.equal(newUserCheck, "emailpassword-true");
+                    assert.equal(urlBeforeSignUp, urlAfterSignUp);
+                });
+            });
+
```
rishabhpoddar, line 434: what happens if you sign in -> get redirected to the home page -> run localStorage.setItem("disableRedirectionAfterSuccessfulSignInUp", "true") -> go back to the auth page. I assume this will behave weirdly in some way? Cause the auth page by default detects if a session exists, and if it does, it tries to redirect, but we are returning null, and then on each render of the auth page, it will try to do this.

CHANGELOG.md
```diff
@@ -5,6 +5,26 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
+## [0.35.8] - 2023-11-24
+
+### Changes
+
+-   `getRedirectionURL` now supports returning `null` to prevent automatic redirection, useful for customizing the behavior after successful sign-in or sign-up.
```
rishabhpoddar, line 12:
```suggestion
-   `getRedirectionURL` now supports returning `null` to prevent automatic redirection, useful for customizing the behavior after successful sign-in or sign-up. Typically, this is used if you want to embed the sign in / up component in a popup and want to just dismiss the popup post login.
```

test/end-to-end/getRedirectionURL.test.js
```diff
@@ -377,5 +375,152 @@ describe("getRedirectionURL Tests", function () {
                 assert.equal(newUserCheck, "thirdpartypasswordless-true");
             });
         });
+
+        describe("No Redirection", function () {
```
rishabhpoddar, line 379: missing:
```
afterEach(function () {
                return screenshotOnFailure(this, browser);
            });
```

test/end-to-end/getRedirectionURL.test.js
```diff
@@ -377,5 +375,152 @@ describe("getRedirectionURL Tests", function () {
...
+                    // wait until network idle to ensure that the page has not been redirected
+                    await page.waitForNetworkIdle();
+
+                    const urlAfterSignUp = await page.url();
+                    const newUserCheck = await page.evaluate(() => localStorage.getItem("isNewUserCheck"));
+                    assert.equal(newUserCheck, "passwordless-true");
+                    assert.equal(urlBeforeSignUp, urlAfterSignUp);
+                });
+            });
+        });
```
rishabhpoddar, line 524: add for third party as well. We want to make sure that the user stays on the consume code screen

test/end-to-end/getRedirectionURL.test.js
```diff
@@ -377,5 +375,152 @@ describe("getRedirectionURL Tests", function () {
...
+                        method: "POST",
+                    }).catch(console.error);
+                    await fetch(`${TEST_SERVER_BASE_URL}/stopst`, {
+                        method: "POST",
+                    }).catch(console.error);
+
+                    return screenshotOnFailure(this, browser);
+                });
+
+                it("should not do any redirection after successful sign up", async function () {
```
rishabhpoddar, line 498: add for passwordless magic link too. Once the link is consumed, we want to make sure that the user stays on that screen - what does the UI there look like after?

rishabhpoddar, line 498: add for passwordless magic link too. Once the link is consumed, we want to make sure that the user stays on that screen - what does the UI there look like after?

test/end-to-end/getRedirectionURL.test.js
```diff
@@ -377,5 +375,152 @@ describe("getRedirectionURL Tests", function () {
...
+                    const device = await getPasswordlessDevice(loginAttemptInfo);
+                    await setInputValues(page, [{ name: "userInputCode", value: device.codes[0].userInputCode }]);
+                    await submitForm(page);
+                    // wait until network idle to ensure that the page has not been redirected
+                    await page.waitForNetworkIdle();
+
+                    const urlAfterSignUp = await page.url();
+                    const newUserCheck = await page.evaluate(() => localStorage.getItem("isNewUserCheck"));
+                    assert.equal(newUserCheck, "passwordless-true");
+                    assert.equal(urlBeforeSignUp, urlAfterSignUp);
```
rishabhpoddar, line 521: at this point, what does the UI on the screen look like - is it just the enter OTP screen UI only?

I ask cause we have [another UI state](https://supertokens.com/docs/passwordless/pre-built-ui/further-reading/passwordless-login#logging-in-via-otp-and-magic-link-simultaneously), which we should not be seeing in this case.

test/end-to-end/getRedirectionURL.test.js
rishabhpoddar: I ask cause we have [another UI state](https://supertokens.com/docs/passwordless/pre-built-ui/further-reading/passwordless-login#logging-in-via-otp-and-magic-link-simultaneously) - does it show this? Or just a blank screen?

rishabhpoddar (changes requested):

lib/ts/types.ts
```diff
@@ -30,7 +30,13 @@ export type GetRedirectionURLContext = {
...
 
 export type ValidationFailureCallback =
-    | (({ userContext, reason }: { userContext: any; reason: any }) => Promise<string | undefined> | string | undefined)
+    | (({
+          userContext,
+          reason,
+      }: {
+          userContext: any;
+          reason: any;
+      }) => Promise<string | undefined | null> | string | undefined | null)
```
porcellus, line 39: In this context does a `null` return value have a separate meaning from `undefined`? 
I'd prefer if this either:
- stayed as `string | undefined` or
- if this was handled identically to `undefined`. (In the above implementation a `null` return value disables the access denied screen, but I don't think that's intentional)

porcellus (changes requested):

test/end-to-end/getRedirectionURL.test.js
anku255: We show the success screen and then no redirection happens.

image

test/end-to-end/getRedirectionURL.test.js
anku255: If a user attempts to access the /auth page while the `getRedirectionURL` function returns `null`, the result will be a blank page without further action.

You are correct that every time the Auth page loads, it will try to redirect. This is not a problem because this only happens per render and returning `null` doesn't cause any more re-renders.

examples/for-tests/src/App.js
anku255: done

test/end-to-end/getRedirectionURL.test.js
anku255: it was removed by the linter. Having a return here makes no difference as the test is skipped completely after calling `this.skip()`. Nevertheless, I have put it back to avoid confusion.

test/end-to-end/getRedirectionURL.test.js
anku255: it was removed by the linter. Having a return here makes no difference as the test is skipped completely after calling `this.skip()`. Nevertheless, I have put it back to avoid confusion.

test/end-to-end/getRedirectionURL.test.js
anku255: This checks that a new user was created (sign up successful) and then no redirection happened. The before and after url would be same in case of a failed sign up as well.

test/end-to-end/getRedirectionURL.test.js
anku255: It is present in the `after` function.

test/end-to-end/getRedirectionURL.test.js
anku255: done

test/end-to-end/getRedirectionURL.test.js
anku255: The magic link page shows a spinner (after successful login)  while the existing enter OTP page redirects to the return value of `getRedirectURL`.

lib/ts/types.ts
anku255: We had a discussion regarding this and the concern was that 

```
onFailureRedirct: () => undefined, showAccessDeniedOnFailure: false
```
would be same as 

`onFailureRedirct: () => null`

To avoid confusion, we didn't wanna have the option to return `null` as well.


However, this is not true. In the first case, the [default redirection](https://github.com/supertokens/supertokens-auth-react/blob/master/lib/ts/recipe/session/recipe.tsx/#L196-L198)  ('/') would still happen while in the latter it won't. We still need to support null to make use cases like `/auth/verify-phone` after login work.

CHANGELOG.md
```diff
@@ -5,11 +5,53 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
-## [0.35.8] - 2023-11-26
+## [0.36.0] - 2023-11-24
```
anku255, line 8:
```suggestion
## [0.36.0] - 2023-12-01
```

CHANGELOG.md
```diff
@@ -5,11 +5,53 @@ All notable changes to this project will be documented in this file.
...
+            if (window.location.pathname !== "/auth/verify-phone") {
+                return "/auth/verify-phone";
+            }
+            return null;
+        },
+    });
+    ```
 
-## [0.35.7] - 2023-11-24
+## [0.35.7] - 2023-11-16
```
anku255, line 54:
```suggestion
## [0.35.7] - 2023-11-24
```

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
```diff
@@ -38,23 +38,33 @@ import type { PasswordlessSignInUpAction, SignInUpState, SignInUpChildProps, Nor
...
                         userContext,
                     });
                     if (hasSession) {
-                        dispatch({ type: "successInAnotherTab" });
+                        clearInterval(checkSessionIntervalHandle);
+                        void SessionRecipe.getInstanceOrThrow().validateGlobalClaimsAndHandleSuccessRedirection(
+                            {
+                                rid: recipeId,
+                                successRedirectContext: {
+                                    action: "SUCCESS",
```
rishabhpoddar, line 58: Passing in `SUCCESS` as the action with `isNewRecipeUser: false` is not a good idea. Cause what if the user is actually a new signed up user? And based on that, the dev might want to take to the dashboard vs an onboarding screen? So in one tab, you would see an onboarding UI, and in another, you may see a dashboard UI (which might also break cause that UI might assume that onboarding is completed)

Since there is no other action available at the moment, I suggest that we remove this thing alltogether and just stick to showing the enter OTP screen / spinner screen - like how it's in other recipes.

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
```diff
@@ -38,23 +38,33 @@ import type { PasswordlessSignInUpAction, SignInUpState, SignInUpChildProps, Nor
 import type { RecipeInterface } from "supertokens-web-js/recipe/passwordless";
 import type { User } from "supertokens-web-js/types";
 
-export const useSuccessInAnotherTabChecker = (
-    state: SignInUpState,
-    dispatch: React.Dispatch<PasswordlessSignInUpAction>,
-    userContext: any
-) => {
+export const useRedirectAfterSuccess = (state: SignInUpState, recipeId: string, userContext: any, history: any) => {
```
rishabhpoddar, line 41: A better name would be `useRedirectAfterSuccessInAnotherTab` - otherwise it's confusing that this is also for the current tab.

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
```diff
@@ -38,23 +38,33 @@ import type { PasswordlessSignInUpAction, SignInUpState, SignInUpChildProps, Nor
...
                     if (hasSession) {
-                        dispatch({ type: "successInAnotherTab" });
+                        clearInterval(checkSessionIntervalHandle);
+                        void SessionRecipe.getInstanceOrThrow().validateGlobalClaimsAndHandleSuccessRedirection(
+                            {
+                                rid: recipeId,
+                                successRedirectContext: {
+                                    action: "SUCCESS",
+                                    isNewRecipeUser: false,
+                                    user: undefined,
```
rishabhpoddar, line 60: Sort of related to this is that the [type of SUCCESS action](https://github.com/supertokens/supertokens-auth-react/blob/master/lib/ts/recipe/authRecipe/types.ts#L29), is strange. The `user` obj can be `undefined` in case of `isNewRecipeUser: false`. This leads to a bad developer experience. 

If there is no place where user is actually undefined (other than this place), and if we remove this "cronjob" then w should change the type of `GetRedirectionURLContext` like:
```
{
          action: "SUCCESS";
          isNewRecipeUser: boolean;
          user: User;
          redirectToPath?: string;
      }
```

lib/ts/utils.ts
```diff
@@ -223,7 +223,15 @@ export function redirectWithFullPageReload(to: string): void {
     if (to.trim() === "") {
         to = "/";
     }
-    WindowHandlerReference.getReferenceOrThrow().windowHandler.location.setHref(to);
+
+    // In certain cases, we invoke `Session.validateGlobalClaimsAndHandleSuccessRedirection` within a setInterval (useRedirectAfterSuccess hook), potentially causing a loop of page reloads if the redirect URL remains unchanged.
+    // To mitigate this, we check whether the current URL matches the redirect URL before initiating redirection.
+    const currentURL = new URL(WindowHandlerReference.getReferenceOrThrow().windowHandler.location.getHref());
+    const destinationURL = new URL(to, currentURL);
```
rishabhpoddar, line 230: why are we passing in `currentURL` as the `base` param?

lib/ts/utils.ts
```diff
@@ -223,7 +223,15 @@ export function redirectWithFullPageReload(to: string): void {
...
+
+    if (destinationURL.href !== currentURL.href) {
```
rishabhpoddar, line 232: does this do redirection if the protocol, domain, path are the same BUT, the query params are not.

test/end-to-end/getRedirectionURL.test.js
rishabhpoddar: The after function runs after all tests. Right now there is only one test in the describe block, but if someone adds another test, it may cause issues? Not sure.

test/end-to-end/getRedirectionURL.test.js
rishabhpoddar: I wonder if we should instead show the login UI as usual in this case as opposed to showing a blank page? If the user does not return `null` from `getRedirectionURL`, then it makes sense to show a blank page other the page might flicker. But if they are returning `null`, it's very weird to just show a blank page.

test/end-to-end/getRedirectionURL.test.js
```diff
@@ -377,5 +377,243 @@ describe("getRedirectionURL Tests", function () {
                 assert.equal(newUserCheck, "thirdpartypasswordless-true");
             });
         });
+
+        describe("No Redirection", function () {
```
rishabhpoddar, line 381: another case to test is that if we are showing the login UI in a popup AND fullpage, then you want to conditionally return `null` from `getRedirectionUrl` function. One way to do this is to use userContext when embedding the SignInUp component, and that same userContext should be available in `getRedirectionUrl`

test/end-to-end/getRedirectionURL.test.js
rishabhpoddar: I don't see a test for this. We need tests for:
- emailpassword
- social login
- password login (OTP or magic link)

So by third party, i meant add the social login flow test as well.

lib/ts/utils.ts
anku255: yes

test/end-to-end/getRedirectionURL.test.js
anku255: fixed it.

test/end-to-end/getRedirectionURL.test.js
anku255: added tests for thirdparty (Auth0) and magic link.

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
anku255: We decided to remove the `user` prop from here as we couldn't guarantee its presence.

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
anku255: Outdated. This function has been removed.

lib/ts/recipe/passwordless/components/features/signInAndUp/index.tsx
anku255: https://github.com/supertokens/supertokens-auth-react/pull/766#discussion_r1413946449

lib/ts/utils.ts
anku255: Update. This change has been removed.

test/end-to-end/getRedirectionURL.test.js
rishabhpoddar: @anku255 and I discussed, and we concluded that we don't need to do `doRedirection` because:
- User does not actually need to pass `shouldRedirectOnSessionExists`
- The requirement to pass userContext to the `SignInUp` component is only if the dev is having a popup AND a full page auth screen - which we feel is a niche use case. In regular just popup case, you don't need to do the userContext thing.

CHANGELOG.md
```diff
@@ -5,6 +5,38 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
+## [0.36.0] - 2023-12-01
+
+### Changes
+
+-   `getRedirectionURL` now supports returning `null` to prevent automatic redirection, useful for customizing the behavior after successful sign-in or sign-up. Typically, this is used if you want to embed the sign in / up component in a popup and want to just dismiss the popup post login.
```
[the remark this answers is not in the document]
anku255: 1. Done
2. It was already there.
3. We do not need to change the signature of `onFailureRedirection`, so no need to update that doc.

lib/ts/recipe/recipeModule/types.ts
```diff
@@ -22,7 +22,10 @@ export type RecipePostAPIHookFunction<Action> = (context: RecipePostAPIHookConte
 export type RecipeOnHandleEventFunction<EventType> = (context: EventType) => void;
 
 export type UserInput<GetRedirectionURLContextType, Action, OnHandleEventContextType> = {
-    getRedirectionURL?: (context: GetRedirectionURLContextType) => Promise<string | undefined>;
+    getRedirectionURL?: (
+        context: GetRedirectionURLContextType,
+        userContext?: any
```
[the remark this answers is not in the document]
rishabhpoddar: Not directly related to this comment, but please also check that the userContext provided to onFailureRedirection is not optional.

lib/ts/recipe/session/recipe.tsx
```diff
@@ -145,9 +166,20 @@ export default class Session extends RecipeModule<unknown, unknown, unknown, Nor
                 userContext,
             });
 
+            if (failureRedirectInfo.redirectPath === null) {
```
rishabhpoddar, line 169: is this required? Cause failureRedirection never returns `null` right?

rishabhpoddar (changes requested):

CHANGELOG.md
```diff
@@ -5,6 +5,52 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
+## [0.36.0] - 2023-12-07
+
+### Changes
+
```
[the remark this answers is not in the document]
anku255: @rishabhpoddar 

1. Already added for `getRedirectionURL`.  Didn't add it for `onFailureRedirection` as it was already present.
2. This didn't change in this PR. Users can already return `undefined`.

test/with-typescript/src/App.tsx
```diff
@@ -459,7 +459,7 @@ function getEmailPasswordConfigs() {
 
         async getRedirectionURL(context: EmailPasswordGetRedirectionURLContext) {
             if (context.action === "SUCCESS") {
-                if (context.isNewRecipeUser && context.user.loginMethods.length === 1) {
```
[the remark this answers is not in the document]
anku255: Updated here - https://github.com/supertokens/supertokens-auth-react/pull/772

lib/ts/recipe/session/sessionAuth.tsx
```diff
@@ -185,11 +186,24 @@ const SessionAuth: React.FC<PropsWithChildren<SessionAuthProps>> = ({ children,
                         overrideGlobalClaimValidators: props.overrideGlobalClaimValidators,
                         userContext,
                     });
+
+                    if (failureRedirectInfo.redirectPath === null) {
```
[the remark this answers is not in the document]
anku255: Good catch! I have removed it now.

lib/ts/recipe/session/recipe.tsx
anku255: Good catch! I have removed it now.

lib/ts/recipe/recipeModule/index.ts
```diff
@@ -28,18 +30,31 @@ export default abstract class RecipeModule<
...
+        }
+
         redirectUrl = appendQueryParamsToURL(redirectUrl, queryParams);
-        return SuperTokens.getInstanceOrThrow().redirectToUrl(redirectUrl, history);
+        return SuperTokens.getInstanceOrThrow().redirectToUrl(redirectUrl, navigate);
     };
 
     // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
-    getRedirectUrl = async (context: GetRedirectionURLContextType): Promise<string> => {
+    getRedirectUrl = async (context: GetRedirectionURLContextType, userContext?: any): Promise<string | null> => {
```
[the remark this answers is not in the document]
anku255: Update here - https://github.com/supertokens/supertokens-auth-react/pull/772

lib/ts/recipe/recipeModule/index.ts
```diff
@@ -28,18 +30,31 @@ export default abstract class RecipeModule<
 > extends BaseRecipeModule<GetRedirectionURLContextType, Action, OnHandleEventContextType, N> {
     redirect = async (
         context: GetRedirectionURLContextType,
-        history?: any,
-        queryParams?: Record<string, string>
+        navigate?: Navigate,
+        queryParams?: Record<string, string>,
+        userContext?: any
```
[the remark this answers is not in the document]
anku255: Update here - https://github.com/supertokens/supertokens-auth-react/pull/772

lib/ts/types.ts
```diff
@@ -329,12 +329,12 @@ export type ThemeBaseProps = {
     styleFromInit?: string;
 };
 
-export type FeatureBaseProps = PropsWithChildren<{
-    /*
-     * History provided by react-router
-     */
-    history?: any;
-}>;
+// eslint-disable-next-line @typescript-eslint/ban-types
```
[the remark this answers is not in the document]
anku255: Fixed.

package.json
```diff
@@ -1,6 +1,6 @@
 {
```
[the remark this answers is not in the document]
anku255: Had a chat with Mihaly. While the % increase may look like a lot, the absolute increase is only a 150-200 bytes. Mihaly and I both think that's okay and not worth investigating further.

PR #772, opened 2023-12-07
Update userContext type and add isNewPrimaryUser in GetRedirectionURLContext

anku255:
## Summary of change

This PR - 

- Updates type of `userContext` from `any` to `Record<string, unknown>`
- Make `userContext` required in internal functions wherever possible
-  Uses `getNormalisedUserContext` to populate `userContext` wherever it was not available/passed
- Add `isNewPrimaryUser` boolean in GetRedirectionURLContext

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

### DIFF
--- a/lib/ts/recipe/emailpassword/components/features/signInAndUp/index.tsx
+++ b/lib/ts/recipe/emailpassword/components/features/signInAndUp/index.tsx
@@ -30,7 +30,7 @@ import Session from "../../../../session/recipe";
 import SignInAndUpTheme from "../../themes/signInAndUp";
 import { defaultTranslationsEmailPassword } from "../../themes/translations";
 
-import type { Navigate, FeatureBaseProps, NormalisedFormField } from "../../../../../types";
+import type { Navigate, FeatureBaseProps, NormalisedFormField, UserContext } from "../../../../../types";
 import type Recipe from "../../../recipe";
 import type { SignInAndUpState } from "../../../types";
 import type {
@@ -158,7 +158,7 @@ export function useChildProps(
             clearError: () => dispatch({ type: "setError", error: undefined }),
             onError: (error: string) => dispatch({ type: "setError", error }),
             onSuccess: onSignInSuccess,
-            forgotPasswordClick: () => recipe.redirect({ action: "RESET_PASSWORD" }, navigate, userContext),
+            forgotPasswordClick: () => recipe.redirect({ action: "RESET_PASSWORD" }, navigate, undefined, userContext),
         };
 
         const signUpForm = {
@@ -230,7 +230,7 @@ const getModifiedRecipeImplementation = (origImpl: RecipeInterface): RecipeInter
 function getThemeSignUpFeatureFormFields(
     formFields: NormalisedFormField[],
     recipe: Recipe,
-    userContext: any
+    userContext: UserContext
 ): FormFieldThemeProps[] {
     const emailPasswordOnly = formFields.length === 2;
     return formFields.map((field) => ({
--- a/lib/ts/recipe/emailpassword/components/features/signInAndUp/index.tsx
+++ b/lib/ts/recipe/emailpassword/components/features/signInAndUp/index.tsx
@@ -117,6 +117,7 @@ export function useChildProps(
                 rid: recipe!.config.recipeId,
                 successRedirectContext: {
                     action: "SUCCESS",
+                    isNewPrimaryUser: false,
                     isNewRecipeUser: false,
                     redirectToPath: getRedirectToPathFromURL(),
                 },
@@ -132,6 +133,7 @@ export function useChildProps(
                 rid: recipe!.config.recipeId,
                 successRedirectContext: {
                     action: "SUCCESS",
+                    isNewPrimaryUser: true,
                     isNewRecipeUser: true,
                     redirectToPath: getRedirectToPathFromURL(),
                 },
--- a/lib/ts/recipe/emailverification/components/features/emailVerification/index.tsx
+++ b/lib/ts/recipe/emailverification/components/features/emailVerification/index.tsx
@@ -29,12 +29,16 @@ import Session from "../../../../session/recipe";
 import EmailVerificationTheme from "../../themes/emailVerification";
 import { defaultTranslationsEmailVerification } from "../../themes/translations";
 
-import type { FeatureBaseProps } from "../../../../../types";
+import type { FeatureBaseProps, UserContext } from "../../../../../types";
 import type Recipe from "../../../recipe";
 import type { ComponentOverrideMap } from "../../../types";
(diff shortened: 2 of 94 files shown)


lib/ts/recipe/emailpassword/components/features/signInAndUp/index.tsx
```diff
@@ -132,6 +133,7 @@ export function useChildProps(
                 rid: recipe!.config.recipeId,
                 successRedirectContext: {
                     action: "SUCCESS",
+                    isNewPrimaryUser: true,
```
rishabhpoddar, line 136: this is not necessary since a sign up can lead to account linking to an existing user. So you need to actually check the number of items in the login method, and if thats 1, only then this is true, else it's false.

rishabhpoddar: Applies to all recipes.

lib/ts/recipe/emailverification/components/features/emailVerification/index.tsx
```diff
@@ -110,7 +114,7 @@ export const EmailVerification: React.FC<Prop> = (props) => {
 
     const signOut = useCallback(async (): Promise<void> => {
         const session = Session.getInstanceOrThrow();
-        await session.signOut(props.userContext);
+        await session.signOut({ userContext: props.userContext ?? userContext });
```
rishabhpoddar, line 117: this type of check for `props.userContext` needs to be done in all places in this file. In fact, [here](https://github.com/supertokens/supertokens-auth-react/blob/b6b563b84ead4e4fb71d7718f33c53e06900568e/lib/ts/recipe/emailverification/components/features/emailVerification/index.tsx#L46) you can just assign userContext like 
```
let userContext = useUserContext()
if (props.userContext !== undefined) {
   userContext = props.userContext
}
```

rishabhpoddar: Make sure you do this everywhere where we use `useUserContext()` (can do a global search for this)

lib/ts/recipe/emailverification/components/features/emailVerification/index.tsx
```diff
@@ -110,7 +114,7 @@ export const EmailVerification: React.FC<Prop> = (props) => {
...
         return redirectToAuthWithHistory();
     }, [props.recipe, redirectToAuthWithHistory]);
```
rishabhpoddar, line 119: shouldn't the dependency array have userContext in it too?

lib/ts/superTokens.tsx
```diff
@@ -195,10 +196,13 @@ export default class SuperTokens {
...
-        let redirectUrl = await this.getRedirectUrl({
-            action: "TO_AUTH",
-            showSignIn: options.show === "signin",
-        });
+        let redirectUrl = await this.getRedirectUrl(
+            {
+                action: "TO_AUTH",
+                showSignIn: options.show === "signin",
+            },
+            getNormalisedUserContext(undefined)
```
rishabhpoddar, line 204: instead of passing in undefined, you should add a userContext input to the `redirectToAuth` function. And of course, the user facing version of this function (in the index.ts) will have the userContext as optional

lib/ts/types.ts
```diff
@@ -358,3 +367,5 @@ export type Navigate =
           goBack: () => void;
       }
     | NavigateFunction;
+
+export type UserContext = Record<string, unknown>;
```
rishabhpoddar, line 371: I think it's better to put `Record<string, any>` cause with `unknown`, devs have to first explicitly check the type of it using `typeof` or something and then use the value. So it can get annoying for them. What do you think?

rishabhpoddar (changes requested):

lib/ts/recipe/thirdpartyemailpassword/components/themes/signInAndUp/index.tsx
```diff
@@ -34,6 +34,7 @@ import { ThemeBase } from "../themeBase";
```
rishabhpoddar, line 34: not related to this file, but you need to modify the ts playground to use the isNewPrimaryUser, and also check that all components that can be embedded by the user should accept a userContext + that getRedirection function also has a userContext as one of the args.

rishabhpoddar (changes requested):

lib/ts/recipe/emailverification/components/features/emailVerification/index.tsx
anku255: This was the only place where `userContext` is also being passed as prop and `useUserContext` has been used.

lib/ts/recipe/emailverification/components/features/emailVerification/index.tsx
anku255: done

lib/ts/recipe/thirdpartyemailpassword/components/themes/signInAndUp/index.tsx
anku255: done

lib/ts/superTokens.tsx
anku255: done

lib/ts/recipe/emailverification/components/features/emailVerification/index.tsx
anku255: done

lib/ts/recipe/emailpassword/components/features/signInAndUp/index.tsx
anku255: done

lib/ts/recipe/session/components/features/accessDeniedScreen/index.tsx
```diff
@@ -6,13 +6,14 @@ import UI from "../../../../../ui";
...
 
-import type { FeatureBaseProps } from "../../../../../types";
+import type { FeatureBaseProps, UserContext } from "../../../../../types";
 import type Recipe from "../../../recipe";
 import type { ComponentOverrideMap } from "../../../types";
 
 const AccessDeniedScreen: React.FC<
     FeatureBaseProps<{
         recipe: Recipe;
+        userContext?: UserContext;
```
rishabhpoddar, line 16: this is not being used anywhere in this file. Is that ok?

rishabhpoddar (changes requested):

lib/ts/recipe/session/components/features/accessDeniedScreen/index.tsx
anku255: Yeah, this is not being used currently but it has been intentionally included here as a reminder that it is passed as a prop when needed.

PR #778, opened 2023-12-12
feat: Remove with-aws-lambda example

anku255:
## Summary of change

This PR removes the `with-aws-lambda` example because we are now using the example in `supertokens-node` (https://github.com/supertokens/supertokens-node/pull/761) instead.

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`


examples/with-aws-lambda/backend/config.mjs
```diff
@@ -0,0 +1,70 @@
+import ThirdPartyEmailPassword from "supertokens-node/recipe/thirdpartyemailpassword";
```
rishabhpoddar, line 1: we should remove everything from this folder and just have a readme that links to the node repo example since that's tested and all + don't want to maintain the same thing here and there

rishabhpoddar (changes requested):

PR #779, opened 2023-12-20
fix: preserve query params when redirectBack is used

anku255:
## Summary of change

Previously, when calling `redirectToAuth` with the `redirectBack` option, query parameters were stripped when redirecting back to the previous page after authentication. This issue has been fixed, and now query parameters are preserved as intended.

## Test Plan

Tested manually extensively. Also, updated an existing test case to verify query params are being preserved. 

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

### DIFF
--- a/lib/ts/utils.ts
+++ b/lib/ts/utils.ts
@@ -80,7 +80,18 @@ export function getRedirectToPathFromURL(): string | undefined {
         try {
             const normalisedURLPath = new NormalisedURLPath(param).getAsStringDangerous();
             const pathQueryParams = param.split("?")[1] !== undefined ? `?${param.split("?")[1]}` : "";
-            return normalisedURLPath + pathQueryParams;
+            const pathWithQueryParams = normalisedURLPath + pathQueryParams;
+
+            // Ensure a leading "/" if `normalisedUrlPath` is empty but `pathWithQueryParams` is not to ensure proper redirection.
+            // Example: "?test=1" will not redirect the user to `/?test=1` if we don't add a leading "/".
+            if (
+                normalisedURLPath.length === 0 &&
+                pathWithQueryParams.length > 0 &&
+                !pathWithQueryParams.startsWith("/")
+            ) {
+                return "/" + pathWithQueryParams;
+            }
+            return pathWithQueryParams;
         } catch {
             return undefined;
         }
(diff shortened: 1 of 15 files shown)


package.json
```diff
@@ -105,6 +105,7 @@
         "build": "rm -rf lib/build && npx rollup -c",
         "watch": "npx rollup -cw",
         "watch-mac": "chokidar lib/ts/* -c 'npx rollup -c'",
+        "watch-mac-2": "chokidar lib/ts/* -c 'npx rollup -c && cp -r lib recipe ui index.d.ts index.js package.json rollup.config.mjs webJsInterfaceSupported.json ../../supertokens-demo/frontend/node_modules/supertokens-auth-react/ && rm -rf ../../supertokens-demo/frontend/node_modules/.cache'",
```
rishabhpoddar, line 108: what is this for?

test/end-to-end/signup.test.js
```diff
@@ -134,10 +134,18 @@ describe("SuperTokens SignUp", function () {
             await page.evaluate(() => window.SuperTokens.redirectToAuth());
```
rishabhpoddar, line 134: it would be good to add / modify tests that actually do redirection backend and see that that preserves query param in the case of:
- email password login
- third party login
- passwordless magic linlk
- passwordless otp
- email verification

There should already be tests for these somewhere where we actually check that it redirects back properly, so you can just modify those to add query params and check that they are preserved.

rishabhpoddar (changes requested):

lib/ts/utils.ts
```diff
@@ -80,7 +80,18 @@ export function getRedirectToPathFromURL(): string | undefined {
...
+            // Ensure a leading "/" if `normalisedUrlPath` is empty but `pathWithQueryParams` is not to ensure proper redirection.
+            // Example: "?test=1" will not redirect the user to `/?test=1` if we don't add a leading "/".
+            if (
+                normalisedURLPath.length === 0 &&
+                pathWithQueryParams.length > 0 &&
+                !pathWithQueryParams.startsWith("/")
+            ) {
+                return "/" + pathWithQueryParams;
+            }
+            return pathWithQueryParams;
```
rishabhpoddar, line 94: this is a bug fix from before right?

rishabhpoddar: If so, and previously, if there were no query params, and the redirectToPath was `/`, we would end up redirecting to `""`? How did that work?

rishabhpoddar (changes requested):

lib/ts/utils.ts
anku255: If the `normalisedURLPath` was empty then it would be considered as if there was no `redirectToPath` and we would end up doing the default `/` redirection. 

The bug presents itself when we preserve query params because now `redirecToPath` exists but without a leading slash.

package.json
anku255: removed it. This was not supposed to be committed.

test/end-to-end/signup.test.js
rishabhpoddar: - redirectBack: false test
- redirect where path is `/`, but has only query params.

rishabhpoddar (changes requested):

test/end-to-end/signup.test.js
anku255: done

rishabhpoddar (approved):

PR #789, opened 2024-01-23
feat: SessionAuth SSR support (initialSessionAuthContext)

sasha240100:
## Summary of change

Adds `initialSessionAuthContext` support for `SessionAuth` component. `initialSessionAuthContext` is an initial state that Session provider uses to create a context.

- Added `initialSessionAuthContext` property to `SessionAuthProps`
- Added `isContextFromSSR` to `LoadedSessionContext` that signals if a session context is provided from a server or computed on the client.
   -  `isContextFromSSR: true`  when uses initial context preloaded from server (SSR). This means `invalidClaims[]` and `accessDeniedValidatorError` are not yet evaluated and the client should not trust the data (as discussed with @porcellus ) 
   -  `isContextFromSSR: false`  when uses a client generated context (ex: after client rerender)

## Related issues

-   https://github.com/supertokens/supertokens-auth-react/issues/790
-   https://github.com/supertokens/supertokens-auth-react/issues/263
-   https://github.com/supertokens/supertokens-node/pull/785

## Test Plan

Tested locally on a NextJS implementation using supertokens. Tested scenarios include:
- Tested `<SessionAuth />` rendering with children that consume context (userId) -> Renders on server side, no flash or hydration issues
- Tested with `requireAuth: false` on server component. ->  `<SessionAuth />`  provides an empty session context and renders children

## Documentation changes

TODO.

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

-   [x] Add `getInitialSessionAuthContext` support to `supertokens-node`
-   [ ] Update documentation to include examples

### DIFF
--- a/lib/ts/recipe/session/sessionAuth.tsx
+++ b/lib/ts/recipe/session/sessionAuth.tsx
@@ -35,6 +35,10 @@ import type { ClaimValidationError } from "supertokens-web-js/recipe/session";
 
 export type SessionAuthProps = {
     /**
+     * Initial context that is rendered on a server side (SSR).
+     */
+    initialSessionAuthContext?: SessionContextType;
+    /**
      * For a detailed explanation please see https://github.com/supertokens/supertokens-auth-react/issues/570
      */
     requireAuth?: boolean;
@@ -67,7 +71,7 @@ const SessionAuth: React.FC<PropsWithChildren<SessionAuthProps>> = ({ children, 
 
     // Reusing the parent context was removed because it caused a redirect loop in an edge case
     // because it'd also reuse the invalid claims part until it loaded.
-    const [context, setContext] = useState<SessionContextType>({ loading: true });
+    const [context, setContext] = useState<SessionContextType>(props.initialSessionAuthContext ?? { loading: true });
 
     const session = useRef<Session>();
 
@@ -101,6 +105,7 @@ const SessionAuth: React.FC<PropsWithChildren<SessionAuthProps>> = ({ children, 
 
         if (sessionExists === false) {
             return {
+                preloaded: false,
                 loading: false,
                 doesSessionExist: false,
                 accessTokenPayload: {},
@@ -128,6 +133,7 @@ const SessionAuth: React.FC<PropsWithChildren<SessionAuthProps>> = ({ children, 
                 throw err;
             }
             return {
+                preloaded: false,
                 loading: false,
                 doesSessionExist: false,
                 accessTokenPayload: {},
@@ -138,6 +144,7 @@ const SessionAuth: React.FC<PropsWithChildren<SessionAuthProps>> = ({ children, 
 
         try {
             return {
+                preloaded: false,
                 loading: false,
                 doesSessionExist: true,
                 invalidClaims,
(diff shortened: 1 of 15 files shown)


lib/ts/recipe/session/sessionAuth.tsx
```diff
@@ -34,6 +34,10 @@ import type { PropsWithChildren } from "react";
 import type { ClaimValidationError } from "supertokens-web-js/recipe/session";
 
 export type SessionAuthProps = {
+    /**
+     * Initial context that is rendered on a server side (SSR).
+     */
+    initialSessionAuthContext?: SessionContextType;
```
porcellus, line 40: since we want to always want to mark this as an SSR context and since there'll be no real claim validation info in it, I'd prefer if those were not part of the input. This should use a separate type

porcellus (changes requested):

lib/ts/recipe/session/sessionAuth.tsx
sasha240100: I'd prefer to keep it the same time as we can't guarantee that `initialSessionAuthContext` is coming from `getInitialSessionAuthContext`, in some case the user might implement a custom logic that will build an `initialSessionAuthContext` with `invalidClaims` for example

recipe/session/index.js
```diff
@@ -13,6 +13,7 @@
  * under the License.
  */
 "use strict";
+"use client"; // Important for NextJS support (SessionAuth is a client component)
```
rishabhpoddar, line 16: we can't do this.

rishabhpoddar: We should create a separate file `/recipe/session/clientcomponents/sessionauth.js`

test/unit/recipe/thirdpartypasswordless/signInUp.test.tsx
```diff
@@ -57,6 +57,7 @@ describe("ThirdPartyPasswordless.SignInAndUp", () => {
         });
```
rishabhpoddar, line 58: we need to add an e2e test that uses SessionAuth without calling the init function and make sure that it works (in the context of SSR)

test/unit/recipe/session/sessionAuth.test.tsx
```diff
@@ -155,6 +156,34 @@ describe("SessionAuth", () => {
 
         result.unmount();
     });
+
+    test("set initial SSR context (initialSessionAuthContext)", async () => {
```
rishabhpoddar, line 160: add some more tests please

lib/ts/recipe/session/sessionAuth.tsx
```diff
@@ -65,9 +74,16 @@ const SessionAuth: React.FC<PropsWithChildren<SessionAuthProps>> = ({ children,
         );
     }
 
+    const initialContext: SessionContextType = props.initialSessionAuthContext
+        ? {
+              ...props.initialSessionAuthContext,
+              invalidClaims: [], // invalidClaims is currently unsupported on server (SSR)
```
rishabhpoddar, line 80:
```suggestion
              invalidClaims: [], // invalidClaims is currently unsupported on server (SSR),
              isContextFromSSR: true,
```

rishabhpoddar (changes requested):

PR #791, opened 2024-01-25
fix: Remove !important from css styles

anku255:
## Summary of change

This PR removes the  `!important` declaration from the CSS styles.

## Test Plan

No further changes were needed after removing the `!important` styles to keep the existing styles. Here is the screenshot after making this change - 

image
image

## Documentation changes

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`


rishabhpoddar:
It seems like you have done manual testing of this for just a few cases? We need a more robust way of testing and making sure nothing broke. In the MFA branch, we have added story book which helps know about changes in the UI. So make this PR on top of the MFA branch and check the storybook thing to see that nothing broke?

porcellus (approved):

PR #799, opened 2024-03-13
fix: example tests

porcellus:
## Summary of change

Renamed `validatorId` to `id`
Small bugfixes/cleanup
Fixing example tests

## Related issues

-   

## Test Plan

Updated tests should cover it

## Documentation changes

Discussed on slack

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`


test/unit/utils.test.tsx
```diff
@@ -176,7 +176,7 @@ describe("Config tests", function () {
             appendQueryParamsToURL("/relative/path?blue=green#with-hash", {
                 foo: "bar",
             }),
-            "/relative/path?blue=green&foo=bar"
+            "/relative/path?blue=green&foo=bar#with-hash"
```
rishabhpoddar, line 179: why is this change here? I don't think this has got to do with MFA PR?

rishabhpoddar (changes requested):

PR #811, opened 2024-04-11
fix: minor UX improvements

porcellus:
## Summary of change

-   We now redirect to the factor chooser screen if the MFA claim validator fails even if there is no available next factor. This will always show an access denied screen, which should help debugging.
-   `clearOnSubmit` now only clears the field value if the request returned an error. This is because the form navigates on success, making clearing the field unnecessary (which can lead to confusing UX if the navigation takes too long).
-   Fixed an issue where `SessionAuth` contents popped in before navigating away if a claim validator failed. Now `SessionAuth` does not set the context before navigation.
-   Made the `name` property optional in custom provider configs
- Added css files to npmignore

## Related issues

-   [node PR](https://github.com/supertokens/supertokens-node/pull/822)

## Test Plan

Updated tests cases should cover this.

Manually tested:
- Claim validator navigation
- Custom provider config crashing/working well with names set in tenant config
- OTP clearing upon error

## Documentation changes

N/A

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`
-   [x] If I added a new login method, I modified the list in `lib/ts/types.ts`
-   [x] If I added a factor id, I modified the list in `lib/ts/recipe/multifactorauth/types.ts`


lib/ts/recipe/session/utils.ts
```diff
@@ -90,3 +92,19 @@ export const getFailureRedirectionInfo = async ({
...
+    const currentUrl = WindowHandlerReference.getReferenceOrThrow().windowHandler.location.getHref();
+    let fullRedirectURL;
+    try {
+        new URL(redirectURL);
+        // if the url is a full, valid url, we can use that
+        fullRedirectURL = redirectURL;
+    } catch {
+        const appInfo = SuperTokens.getInstanceOrThrow().appInfo;
+        // otherwise we prepend the websiteDomain
+        fullRedirectURL = `${appInfo.websiteDomain.getAsStringDangerous()}${redirectURL}`;
```
rishabhpoddar, line 106: why should normalise the redirecturl using normalise path? Cause if the dev gives something like `second-factor` instead of `/second-factor`, then the resulting fullRedirectUrl will always be bad, and never match the current url anyway. But im not sure if normalising will drop query params / fragment etc. 

Also, maybe it's worth checking if the resulting fullRedirectUrl is a valid url and if it's not, then what do we do? Maybe we should throw an error in that case cause it's a bad developer config.

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "supertokens-auth-react",
-    "version": "0.39.1",
+    "version": "0.39.2",
```
rishabhpoddar, line 3: consider making this a breaking change?

rishabhpoddar (changes requested):

lib/ts/recipe/session/utils.ts
porcellus: We are not normalising the redirectUrl because we are not  normalizing it when we are "executing" the redirection either. If the dev gives `second-factor` (let's say on `localhost:3000/dashboard`), then the redirect will go to `localhost:3000/dashboard/second-factor`. 
This makes sense: the dev gave a relative redirection url, which should always redirect you. This means that the behaviour is as expected (it'll not match the current url, the helper returns false). I can add this as an explicit case if you prefer.

package.json
porcellus: I'm fine with making this a breaking change.

Issue #841, opened 2024-07-30
Add a way to end-to-end test OAuth2 flows


porcellus:
- Integrate OAuth2Provider in the test app frontend and backend
  - Expose a test api to handle creating clients in e2e tests
- Integrate OAuth2Client into our test app backend
- Add a separate route in the test app frontend that has a single `Login via SuperTokens` (akin to what we have for TP login)
  - This should get the authorization url from OAuth2Client and redirect to it on button press
  - This should take the clientId from a queryparam
- Add a separate route in the test app frontend that handles the callbacks we get after starting a login process via the above route
  - This should call the signin API of the OAuth2Client
  - It should display session information after that happens or an error if it fails

PR #843, opened 2024-08-01
test: Add e2e test for OAuth2

anku255:
## Summary of change

This PR adds a e2e test for OAuth2 using a generic react OAuth2 library. 

## Related issues

-   https://github.com/supertokens/supertokens-auth-react/issues/841

## Test Plan

## Documentation changes

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

-   [ ] Item1
-   [ ] Item2

### DIFF
--- a/test/end-to-end/oauth2provider.test.js
+++ b/test/end-to-end/oauth2provider.test.js
@@ -1,0 +1,187 @@
+/* Copyright (c) 2021, VRAI Labs and/or its affiliates. All rights reserved.
+ *
+ * This software is licensed under the Apache License, Version 2.0 (the
+ * "License") as published by the Apache Software Foundation.
+ *
+ * You may not use this file except in compliance with the License. You may
+ * obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+/*
+ * Imports
+ */
+
+import assert from "assert";
+import puppeteer from "puppeteer";
+import {
+    clearBrowserCookiesWithoutAffectingConsole,
+    toggleSignInSignUp,
+    screenshotOnFailure,
+    backendBeforeEach,
+    waitForUrl,
+    createOAuth2Client,
+    setOAuth2ClientIdInStorage,
+    removeOAuth2ClientIdFromStorage,
+    getOAuth2LoginButton,
+    getOAuth2LogoutButton,
+    getOAuth2TokenData,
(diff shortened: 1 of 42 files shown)


test/end-to-end/oauth2provider.test.js
```diff
@@ -0,0 +1,187 @@
...
+        }).catch(console.error);
+
+        browser = await puppeteer.launch({
+            args: ["--no-sandbox", "--disable-setuid-sandbox"],
+            headless: true,
+        });
+    });
+
+    after(async function () {
+        await browser.close();
```
porcellus, line 71: Please check that this "passes" (is properly skipped) when running with react16: I have a suspicion the after fn runs anyway and then `browser.close` fails.

test/helpers.js
```diff
@@ -1105,3 +1105,34 @@ export async function expectErrorThrown(page, cb) {
...
+export async function createOAuth2Client(input) {
+    const resp = await fetch(`${TEST_APPLICATION_SERVER_BASE_URL}/test/create-oauth2-client`, {
+        method: "POST",
+        headers: { "content-type": "application/json" },
+        body: JSON.stringify(input),
+    });
+    return await resp.json();
+}
+
+export async function setOAuth2ClientIdInStorage(page, clientId) {
```
porcellus, line 1118: Please add a note on why we are storing this in localstorage (I suspect it's because we need it in the callback)

test/end-to-end/oauth2provider.test.js
```diff
@@ -0,0 +1,187 @@
...
+
+    describe("Generic OAuth2 Client Library", function () {
+        afterEach(async function () {
+            await removeOAuth2ClientIdFromStorage(page);
+        });
+
+        it("should successfully complete the OAuth2 flow", async function () {
+            const { client } = await createOAuth2Client({
+                scope: "offline_access profile openid email",
+                redirectUris: [`${TEST_CLIENT_BASE_URL}/oauth2/callback`],
```
porcellus, line 105: Please add a note somewhere in this file about using the same URL (and why we do not need to have a separate URL for the flow to work right)

porcellus (changes requested):

test/end-to-end/oauth2provider.test.js
anku255: Not sure what you meant. We use different url but the component/page is the same. I have left a comment about that in the OAuth2Page component.

test/helpers.js
anku255: Yes, I have added a comment.

test/end-to-end/oauth2provider.test.js
anku255: You were right. Used a flag to prevent running `after` block if it was skipped.

test/end-to-end/oauth2provider.test.js
porcellus: Ah right, I meant why we are using the same domain.

PR #846, opened 2024-08-08
fix: improve error messages when the authpage has issues covering all factors

porcellus:
## Summary of change

Added a more detailed error message in cases where we have more specific info.

## Related issues

-   [Discord convo](https://discord.com/channels/603466164219281420/1269943257060737085)

## Documentation changes

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

-   [ ] Item1
-   [ ] Item2

### DIFF
--- a/lib/ts/recipe/authRecipe/components/feature/authPage/authPage.tsx
+++ b/lib/ts/recipe/authRecipe/components/feature/authPage/authPage.tsx
@@ -444,7 +444,26 @@ async function buildAndSetChildProps(
     const selectedComponents = selectComponentsToCoverAllFirstFactors(partialAuthComps, firstFactors);
 
     if (selectedComponents === undefined) {
-        throw new Error("Couldn't cover all first factors");
+        const availableFactors = new Set();
+        for (const comp of partialAuthComps) {
+            for (const id of comp.factorIds) {
+                availableFactors.add(id);
+            }
+        }
+        const source =
+            factorListState !== undefined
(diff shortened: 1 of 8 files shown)


lib/ts/recipe/authRecipe/components/feature/authPage/authPage.tsx
```diff
@@ -368,6 +372,11 @@ async function buildAndSetChildProps(
                 (!SuperTokens.usesDynamicLoginMethods || loadedDynamicLoginMethods!.thirdparty.providers.length === 0)
             ) {
                 firstFactors = firstFactors.filter((f) => f !== FactorIds.THIRDPARTY);
```
rishabhpoddar, line 374: the user was seeing a different error message like frontend doesn't cover all factors or something. I think wherever that message is, that message can also be improved to state exactly which factors are coming from the backend, and which ones are on the frontend, and which ones are missing?

rishabhpoddar (changes requested):

lib/ts/recipe/authRecipe/components/feature/authPage/authPage.tsx
porcellus: That case should be caught earlier by what I added (so we can show a more specific error message), but I also tried to extend the existing one with what the first factors are, where we got the list from and what we have available.

PR #877, opened 2025-02-04
Next.js SSR Improvements

bcbogdan:


lib/ts/nextjs/ssr.ts
```diff
@@ -0,0 +1,218 @@
...
+                return session as LoadedSessionContext;
+            default:
+                // This is here just to prevent typescript from complaining
+                // about the function not returning a value
+                throw new Error(`Unknown state: ${state}`);
+        }
+    }
+
+    // TODO: This method is isolated atm to make it easier to test and debug
+    // In the end it should be merged in the getSSRSession function
```
porcellus, line 81: Why do we need to merge it into getSSRSession? At first glance this works well as both the usecase and the impl is different enough.

lib/ts/nextjs/ssr.ts
```diff
@@ -0,0 +1,218 @@
...
+    static async getServerSidePropsSession(cookies: CookiesObject): Promise<GetServerSidePropsReturnValue> {
+        const redirectTo = "/";
+
+        const authPagePath = `${getAuthPagePath()}`;
+        const refreshLocation = `/api/auth/session/refresh?redirectTo=${redirectTo}`;
```
porcellus, line 86: `getRefreshLocation`?

lib/ts/nextjs/ssr.ts
```diff
@@ -0,0 +1,218 @@
...
+        }
+        return { payload: parsedToken.up, ate: parsedToken.ate, isValid: true };
+    } catch (err) {
+        logDebugMessage(`Error while parsing front token: ${err}`);
+        return { isValid: false };
+    }
+}
+
+// TODO:
+// - Do we need to check the token version and handle ERR_JWKS_MULTIPLE_MATCHING_KEYS like in the node SDK?
```
porcellus, line 190: I don't think we do, that only comes up with really old tokens IIRC

lib/ts/nextjs/ssr.ts
```diff
@@ -0,0 +1,218 @@
...
+async function parseAccessToken(
+    token: string
+): Promise<{ isValid: true; payload: AccessTokenPayload["up"] } | { isValid: false }> {
+    const appInfo = SuperTokensNextjsSSRAPIWrapper.getConfigOrThrow().appInfo;
+    const jwksUrl = new URL(`${appInfo.apiBasePath}/jwt/jwks.json`, appInfo.apiDomain);
+    const JWKS = jose.createRemoteJWKSet(jwksUrl);
```
porcellus, line 196: This should be cached/moved into a global location.

lib/ts/nextjs/middleware.ts
```diff
@@ -0,0 +1,204 @@
...
+    const redirectTo = requestUrl.searchParams.get(REDIRECT_PATH_PARAM_NAME) || "/";
+    try {
+        const tokens = await fetchNewTokens(refreshToken);
+        const hasRequiredCookies = tokens.accessToken.cookie && tokens.refreshToken.cookie && tokens.frontToken.cookie;
+        const hasRequiredHeaders = tokens.accessToken.header && tokens.refreshToken.header && tokens.frontToken.header;
+        if (!hasRequiredCookies && !hasRequiredHeaders) {
+            logDebugMessage("Missing tokens from refresh response");
+            return redirectToAuthPage(request);
+        }
+        const redirectUrl = new URL(redirectTo, request.url);
```
porcellus, line 50: Do we check if redirectTo is relative?

lib/ts/nextjs/middleware.ts
```diff
@@ -0,0 +1,204 @@
...
+        const finalResponse = new Response(null, {
+            status: 307,
```
porcellus, line 52: why 307? (and maybe add the explanation as a comment)

porcellus: connected to this, does this handle retrying form posts?

lib/ts/nextjs/middleware.ts
```diff
@@ -0,0 +1,204 @@
...
+            headers: {
+                Location: redirectUrl.toString(),
+            },
+        });
+        finalResponse.headers.append(FRONT_TOKEN_HEADER_NAME, tokens.frontToken.header as string);
```
porcellus, line 57: I don't think we need/want this.

lib/ts/nextjs/middleware.ts
```diff
@@ -0,0 +1,204 @@
...
+        if (hasRequiredCookies) {
+            finalResponse.headers.append("set-cookie", `${tokens.accessToken.cookie as string}`);
+            finalResponse.headers.append("set-cookie", `${tokens.refreshToken.cookie as string}`);
+            finalResponse.headers.append("set-cookie", `${tokens.frontToken.cookie as string}`);
+        }
+        if (hasRequiredHeaders) {
```
porcellus, line 63: these should still be set as cookies.

lib/ts/nextjs/middleware.ts
```diff
@@ -0,0 +1,204 @@
...
+            finalResponse.headers.append(REFRESH_TOKEN_HEADER_NAME, tokens.refreshToken.header as string);
+            finalResponse.headers.append(ACCESS_TOKEN_HEADER_NAME, tokens.accessToken.header as string);
```
porcellus, line 65: we still want to save the front-token

lib/ts/nextjs/middleware.ts
```diff
@@ -0,0 +1,204 @@
...
+    try {
+        const signOutResponse = await fetch(signOutURL, {
+            method: "POST",
+            headers: {
+                "Content-Type": "application/json",
+            },
+            credentials: "include",
+        });
+
+        if (signOutResponse.status !== 200) {
```
porcellus, line 103: This can be a 401, which would require a retry

lib/ts/nextjs/middleware.ts
```diff
@@ -0,0 +1,204 @@
...
+        logDebugMessage("Error revoking session");
+        logDebugMessage(err as unknown as string);
+        return;
+    }
+}
+
+function redirectToAuthPage(request: Request): Response {
+    const authPagePath = AppInfo.websiteBasePath || "/auth";
+    const redirectUrl = new URL(authPagePath, request.url);
+    redirectUrl.searchParams.set(FORCE_LOGOUT_PATH_PARAM_NAME, "true");
```
porcellus, line 126: where is this used?

lib/ts/nextjs/middleware.ts
```diff
@@ -0,0 +1,204 @@
...
+    refreshToken: SuperTokensRequestToken;
+    frontToken: SuperTokensRequestToken;
+    antiCsrfToken: SuperTokensRequestToken;
+}> {
+    const refreshApiURL = new URL(`${AppInfo.apiBasePath}/session/refresh`, AppInfo.apiDomain);
+    const refreshResponse = await fetch(refreshApiURL, {
+        method: "POST",
+        headers: {
+            "Content-Type": "application/json",
+            Cookie: `sRefreshToken=${currentRefreshToken}`,
```
porcellus, line 147: this isn't always allowed, we should set it in a cookie only if it came from a cookie.

lib/ts/nextjs/middleware.ts
```diff
@@ -0,0 +1,204 @@
...
+        cookie: null,
+    };
+    const antiCsrfToken: SuperTokensRequestToken = {
+        header: refreshResponse.headers.get(ANTI_CSRF_TOKEN_HEADER_NAME),
+        cookie: null,
+    };
+
+    // getSetCookie was added in node 18 and our build target is ES5
+    // This should not a problem here since the function runs in the Vercel edge runtime environment
+    // @ts-expect-error TS(2339): Property 'getSetCookie' does not exist on type 'Headers'.
```
porcellus, line 171: As a TODO: let's review if we could raise our build-target

lib/ts/recipe/session/utils.ts
```diff
@@ -113,3 +113,5 @@ export function validateAndCompareOnFailureRedirectionURLToCurrent(redirectURL:
 
     return currentUrl === fullRedirectURL;
 }
+
+export function getSessionForSSR() {}
```
porcellus, line 117: I think this can be removed.

lib/ts/recipe/session/types.ts
```diff
@@ -75,3 +75,28 @@ export type AccessDeniedThemeProps = {
 export type ComponentOverrideMap = {
     SessionAccessDenied_Override?: ComponentOverride<typeof AccessDeniedScreenTheme>;
 };
+
+export type AccessTokenPayload = {
```
porcellus, line 79: I think this can be removed.

lib/ts/recipe/session/ssr.ts
```diff
@@ -0,0 +1,154 @@
+import * as jose from "jose";
```
porcellus, line 1: I think this can be removed.

lib/ts/scratch.ts
```diff
@@ -0,0 +1,205 @@
+import * as jose from "jose";
```
porcellus, line 1: You might want to remove this or move it into the with-typescript folder (that's what we use to "test" typings in ci/scripts as well as a scratch file)

porcellus (changes requested):

examples/with-next-ssr-app-directory/app/components/home.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+export async function HomePage() {
+    const cookiesStore = await cookies();
+    const headersStore = await headers();
+    const session = await getSSRSession(cookiesStore, redirect);
+    console.log(session);
+
+    /**
+     * SessionAuthForNextJS will handle proper redirection for the user based on the different session states.
+     * It will redirect to the login page if the session does not exist etc.
```
porcellus, line 23: I'd say it handles redirections on the client side.

lib/ts/nextjs/ssr.ts
bcbogdan: resolved

lib/ts/nextjs/middleware.ts
bcbogdan: I don't think we need to. Since we are the ones setting it in the SSR function

lib/ts/nextjs/middleware.ts
bcbogdan: resolved

lib/ts/nextjs/middleware.ts
bcbogdan: resolved

lib/ts/nextjs/middleware.ts
bcbogdan: resolved

lib/ts/nextjs/middleware.ts
bcbogdan: resolved

lib/ts/nextjs/middleware.ts
bcbogdan: In the following scenarios:
- if the refresh token cannot be found
- if the refresh request fails
- if, due to an unforeseen error, the ssr get session operation ends up in a loop

lib/ts/nextjs/middleware.ts
bcbogdan: resolved

lib/ts/nextjs/middleware.ts
bcbogdan: resolved

lib/ts/nextjs/middleware.ts
```diff
@@ -0,0 +1,323 @@
...
+    AppInfo = config.appInfo;
+    if (config.enableDebugLogs) {
+        enableLogging();
+    }
+
+    // Cancel the refresh cycle if an unforseen state is encountered
+    const redirectAttemptNumber = getRedirectAttemptNumber(request);
+    if (redirectAttemptNumber > REDIRECT_ATTEMPT_MAX_COUNT) {
+        return redirectToAuthPage(request);
+    }
```
porcellus, line 81: This is so good... we may want to add this in backend SDKs in general.

porcellus (changes requested):

lib/ts/nextjs/middleware.ts
```diff
@@ -0,0 +1,323 @@
...
+        config.appInfo.websiteDomain
+    );
+
+    return async (request: Request) => {
+        const requestUrl = new URL(request.url);
+        if (requestUrl.pathname.startsWith(SESSION_REFRESH_API_PATH) && request.method === "GET") {
+            return refreshSession(config, request);
+        }
+
+        if (requestUrl.pathname.startsWith("/api") && usesTheNextjsApiAsTheAuthenticationServer) {
```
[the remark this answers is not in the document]
bcbogdan: Not in the context of nextjs. All api functions are served through the `/api` path.
The first check can be translated to: `isApiRequest`.

lib/ts/nextjs/constants.ts
```diff
@@ -0,0 +1,14 @@
...
+export const REFRESH_TOKEN_COOKIE_NAME = "sRefreshToken";
+export const REFRESH_TOKEN_HEADER_NAME = "st-refresh-token";
+export const ANTI_CSRF_TOKEN_COOKIE_NAME = "sAntiCsrf";
+export const ANTI_CSRF_TOKEN_HEADER_NAME = "anti-csrf";
+export const REDIRECT_ATTEMPT_MAX_COUNT = 5;
+export const REDIRECT_ATTEMPT_COUNT_COOKIE_NAME = "sSsrSessionRefreshAttempt";
+export const CURRENT_PATH_COOKIE_NAME = "sCurrentPath";
+export const FORCE_LOGOUT_PATH_PARAM_NAME = "forceLogout";
+export const REDIRECT_PATH_PARAM_NAME = "stRedirectTo";
+export const SESSION_REFRESH_API_PATH = "/api/auth/session/refresh";
```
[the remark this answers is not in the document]
bcbogdan: It is. Maybe the naming is wrong here. That's the path that we _expose_ with this custom middleware. It gets called during SSR with a `GET` request. From there we call the actual refresh path (determined based on the `appInfo` data) to get the new tokens.

lib/ts/nextjs/middleware.ts
```diff
@@ -0,0 +1,323 @@
...
+    FRONT_TOKEN_COOKIE_NAME,
+    ANTI_CSRF_TOKEN_COOKIE_NAME,
+    CURRENT_PATH_COOKIE_NAME,
+    SESSION_REFRESH_API_PATH,
+} from "./constants";
+
+import type { ApiRequestMiddleware, SuperTokensNextjsConfig } from "./types";
+
+let AppInfo: SuperTokensNextjsConfig["appInfo"];
+const DEFAULT_API_BASE_PATH = "/api/auth";
```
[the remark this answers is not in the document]
bcbogdan: The `apiBasePath` property for nextjs is `/api/auth`. 
In the `normaliseInputAppInfoOrThrowError` function we use `/auth/` as the default base path. If we want to use that function in this context I'll have to add a new optional parameter to it. Since we can't determine if it's a nextjs app just from the `AppInfo`.

from this
```ts
function normaliseInputAppInfoOrThrowError(appInfo: AppInfoUserInput): NormalisedAppInfo 
```

to this
```ts
function normaliseInputAppInfoOrThrowError(appInfo: AppInfoUserInput, defaultApiBasePath?: string): NormalisedAppInfo 
```

Doesn't seem like the best solution imo.

lib/ts/nextjs/middleware.ts
```diff
@@ -0,0 +1,323 @@
...
+    }
+    const response = new Response(null, {});
+
+    try {
+        const accessToken =
+            getCookie(request, ACCESS_TOKEN_COOKIE_NAME) || request.headers.get(ACCESS_TOKEN_HEADER_NAME);
+        if (!accessToken) {
+            throw new Error("No access token found in the request");
+        }
+        const signOutURL = new URL(`${AppInfo.apiBasePath}/signout`, AppInfo.apiDomain);
```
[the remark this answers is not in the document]
bcbogdan: In that case doesn't that mean that the session is already revoked? As in, we don't need to concern ourselves with invalidating it.

lib/ts/nextjs/middleware.ts
```diff
@@ -0,0 +1,323 @@
...
+                Cookie: `${ACCESS_TOKEN_COOKIE_NAME}=${accessToken}`,
+            },
+            credentials: "include",
+        });
+    } catch (err) {
+        logDebugMessage("Error during the sign out attempt");
+        logDebugMessage(err as unknown as string);
+    }
+
+    response.headers.set("x-middleware-next", "1");
```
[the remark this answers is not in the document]
bcbogdan: it's a nextjs specific header. It signals that the response can be passed to the next middleware (if it exists). It's the equivalent of doing `NextResponse.next()`

lib/ts/nextjs/middleware.ts
```diff
@@ -0,0 +1,323 @@
...
+    response.headers.delete("set-cookie");
```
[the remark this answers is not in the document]
bcbogdan: How can we do this from inside the middleware? Since we don't have access to browser APIs?.

lib/ts/nextjs/middleware.ts
```diff
@@ -0,0 +1,323 @@
...
+        refreshRequestHeaders.Authorization = `Bearer ${headerRefreshToken}`;
+    }
+
+    const refreshResponse = await fetch(refreshApiURL, {
+        method: "POST",
+        headers: refreshRequestHeaders,
+        credentials: "include",
+    });
+
+    if (refreshResponse.status === 401 && attemptNumber <= MAX_REFRESH_ATTEMPTS) {
```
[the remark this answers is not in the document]
bcbogdan: resolved

lib/ts/nextjs/middleware.ts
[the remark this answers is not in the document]
porcellus: In my quick test, that doesn't seem to be true. Are there any docs stating it, or is it just a convention?

lib/ts/nextjs/constants.ts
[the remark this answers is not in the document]
porcellus: Shouldn't it take apiBasePath into account?

lib/ts/nextjs/middleware.ts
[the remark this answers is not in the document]
porcellus: In the generated example app the appinfo looks like this:
```
export const appInfo = {
    appName: "SuperTokens Next.js demo app",
    apiDomain: "http://localhost:3000",
    websiteDomain: "http://localhost:3000",
    apiBasePath: "/api/auth",
    websiteBasePath: "/auth",
};
```
Which means that the apiBasePath is set explicitly. I think the default is always `/auth`

lib/ts/nextjs/middleware.ts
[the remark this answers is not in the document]
porcellus: No that could be only the access token expiring

lib/ts/nextjs/middleware.ts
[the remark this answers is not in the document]
porcellus: We could be clearing the cookies, right?

lib/ts/nextjs/middleware.ts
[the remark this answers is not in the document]
bcbogdan: Added an optional config function that determines if a request is an API request. By default we check for the "/api" path, but users can add any type of logic based on their context. This way we can make it explicit for them that they need to add the property if they have a custom setup.

lib/ts/nextjs/constants.ts
```diff
@@ -0,0 +1,14 @@
+export const ACCESS_TOKEN_COOKIE_NAME = "sAccessToken";
+export const ACCESS_TOKEN_HEADER_NAME = "st-access-token";
+export const FRONT_TOKEN_COOKIE_NAME = "sFrontToken";
+export const FRONT_TOKEN_HEADER_NAME = "front-token";
+export const REFRESH_TOKEN_COOKIE_NAME = "sRefreshToken";
+export const REFRESH_TOKEN_HEADER_NAME = "st-refresh-token";
+export const ANTI_CSRF_TOKEN_COOKIE_NAME = "sAntiCsrf";
+export const ANTI_CSRF_TOKEN_HEADER_NAME = "anti-csrf";
+export const REDIRECT_ATTEMPT_MAX_COUNT = 5;
+export const REDIRECT_ATTEMPT_COUNT_COOKIE_NAME = "sSsrSessionRefreshAttempt";
```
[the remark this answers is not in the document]
bcbogdan: Resolved

lib/ts/nextjs/constants.ts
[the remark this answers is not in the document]
bcbogdan: resolved

lib/ts/nextjs/middleware.ts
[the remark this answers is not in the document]
bcbogdan: resolved

package.json
```diff
@@ -91,6 +91,7 @@
     },
     "dependencies": {
         "intl-tel-input": "^17.0.19",
+        "jose": "^6.0.8",
```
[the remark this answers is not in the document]
bcbogdan: resolved

lib/ts/nextjs/types.ts
```diff
@@ -0,0 +1,77 @@
...
+    appInfo: AppInfoUserInput;
+    enableDebugLogs?: boolean;
+};
+
+export type SuperTokensRequestToken = {
+    header: string | null;
+    cookie: string | null;
+};
+
+export type AccessTokenPayload = {
```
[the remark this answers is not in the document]
bcbogdan: resolved

lib/ts/nextjs/middleware.ts
bcbogdan: Renamed this into `confirmAuthenticationAndCallServerAction` since `authenticateServerAction` might lead users to wrong conclusions. It just checks the authentication state (and retries if required). We can't do the flow where we call the server action first and check for a 401 since serve actions can't return 401 responses.

If they want to access the authentication session object inside the server action they need to call the other function, 'getServerActionSession`.

lib/ts/nextjs/middleware.ts
```diff
@@ -321,3 +334,22 @@ function compareUrlHost(firstUrl: string, secondUrl: string): boolean {
...
+function getRefreshAPIPath(): string {
+    const apiPath = AppInfo.apiBasePath.getAsStringDangerous();
+    return `${apiPath}/session/refresh`;
+}
+
+function defaultIsApiRequest(request: Request): boolean {
+    const requestUrl = new URL(request.url);
+    const refreshPath = getRefreshAPIPath();
+    return requestUrl.pathname.startsWith(refreshPath);
+}
```
porcellus, line 355: I think this isn't right, or the name is misleading.

lib/ts/nextjs/ssr.ts
```diff
@@ -135,14 +125,13 @@ export default class SuperTokensNextjsSSRAPIWrapper {
...
+     * Ensures that a server action is called by an authenticated user
      * If the session does not exist/user is not authenticated, it will automatically redirect to the login page
      * The function is meant to run on the client, before calling the actual server action
-     * @param action - A server action that takes the session context as its first parameter
+     * @param action - A server action that will get called after the authentication state is confirmed
      * @returns The server action return value
      **/
-    static async authenticateServerAction<T extends (session: LoadedSessionContext) => Promise<K>, K>(action: T) {
-        let loadedSessionContext: LoadedSessionContext | undefined = undefined;
+    static async confirmAuthenticationAndCallServerAction<T extends () => Promise<K>, K>(action: T) {
```
porcellus, line 134: I think a shorter name would be better. 
```suggestion
    static async ensureSessionAndCall<T extends () => Promise<K>, K>(action: T) {
```

lib/ts/nextjs/ssr.ts
```diff
@@ -211,18 +201,22 @@ export const init = SuperTokensNextjsSSRAPIWrapper.init;
...
+function getRefreshAPIPath(): string {
+    const apiPath = SuperTokensNextjsSSRAPIWrapper.getConfigOrThrow().appInfo.apiBasePath || DEFAULT_API_PATH;
+    return `${apiPath}/session/refresh`;
+}
+
 function getRefreshLocation(redirectPath: string): string {
-    return `${SESSION_REFRESH_API_PATH}?${REDIRECT_PATH_PARAM_NAME}=${redirectPath}`;
+    const refreshAPIPath = getRefreshAPIPath();
+    return `${refreshAPIPath}?${REDIRECT_PATH_PARAM_NAME}=${redirectPath}`;
 }
```
porcellus, line 220: I think the first name is also present in the middleware file, you could indicate in the name that one path is meant for the frontend/as a full page redirect.

lib/ts/utils.ts
```diff
@@ -556,6 +559,121 @@ export function useRethrowInRender() {
...
+    if (!kid) {
+        throw new Error("JWT header missing kid (Key ID)");
+    }
+
+    const alg = header.alg as string;
+    if (!alg) {
+        throw new Error("JWT header missing alg (Algorithm)");
+    }
+
+    const jwksResponse = await fetch(jwksUrl);
```
porcellus, line 586: This will reload the jwks on every action call, could we cache this somehow? Also, I'm not necessarily opposed to using jose if it doesn't mess with our bundle sizes.

porcellus (changes requested):

lib/ts/nextjs/middleware.ts
```diff
@@ -0,0 +1,438 @@
...
+        }
+        if (header.includes(REFRESH_TOKEN_COOKIE_NAME)) {
+            cookies.push(header);
+        }
+        if (header.includes(ANTI_CSRF_TOKEN_COOKIE_NAME)) {
+            cookies.push(header);
+        }
+    }
+
+    if (!cookies.some((cookie) => cookie.includes(ACCESS_TOKEN_COOKIE_NAME))) {
```
porcellus, line 288: Checking like this feels a bit strange - shouldn't we just check if the original refresh response had the response header for header-based sessions or not?

porcellus: Also, keep in mind that there are configurations (`exposeAccessTokenToFrontendInHeaderBasedAuth`) that'd mean both the cookie and the response header are there.

lib/ts/nextjs/middleware.ts
```diff
@@ -0,0 +1,438 @@
...
+        cookies.push(`${ACCESS_TOKEN_HEADER_NAME}=${tokens.accessToken}; Path=/; HttpOnly; SameSite=Lax`);
+    }
+    if (!cookies.some((cookie) => cookie.includes(REFRESH_TOKEN_COOKIE_NAME))) {
+        cookies.push(
+            `${REFRESH_TOKEN_HEADER_NAME}=${tokens.refreshToken}; Path=/api/auth/session/refresh; HttpOnly; SameSite=Lax`
```
porcellus, line 293: I'm not sure this `Path` is correct. This is intended to be a frontend cookie; I believe it's set without a path.

lib/ts/nextjs/types.ts
```diff
@@ -0,0 +1,86 @@
...
+        antiCsrfToken: string | null;
+        iss: string;
+        "st-role": {
+            v: string;
+            t: number[];
+        };
+        "st-perm": {
+            v: string[];
+            t: number;
+        };
```
porcellus, line 51: there can be other (even custom) claims in there (for example `st-ev`), do we need to account for that?

lib/ts/nextjs/constants.ts
```diff
@@ -0,0 +1,16 @@
+export const ACCESS_TOKEN_COOKIE_NAME = "sAccessToken";
+export const ACCESS_TOKEN_HEADER_NAME = "st-access-token";
```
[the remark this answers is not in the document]
porcellus: I guess we reuse the same name for getting the header, we may want to split them into two consts, as it's confusing to see this being used for setting cookies.

lib/ts/nextjs/middleware.ts
```diff
@@ -0,0 +1,438 @@
...
+        // Used to redirect the user to the correct path after login/refresh
+        // https://github.com/vercel/next.js/issues/43704#issuecomment-2090798307
+        // TL;DR: You can not access pathname in SSR and requests that originate from redirect()
+        const response = next();
+        if (!isInternalPath(requestUrl.pathname)) {
+            response.headers.append(
+                "set-cookie",
+                `${CURRENT_PATH_COOKIE_NAME}=${requestUrl.pathname}; Path=/; HttpOnly; SameSite=Strict`
+            );
+        }
```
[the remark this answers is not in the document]
bcbogdan: The request that gets handled last will be the source of truth. Though I don't think this will affect us since the cookie gets set only when the middleware handles pages.
I've added a max age of a minute

lib/ts/nextjs/constants.ts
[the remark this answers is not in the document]
bcbogdan: resolved

lib/ts/nextjs/middleware.ts
bcbogdan: As far as I can tell, from the inspecting existing refresh cookies and from what [the node sdk](https://github.com/supertokens/supertokens-node/blob/master/lib/ts/recipe/session/cookieAndHeaders.ts#L175) sends to the client a separate path is set on the cookie.

A potential bug here is that I'm not taking into account the AppInfo when setting the path. I fixed with the latest change.

lib/ts/nextjs/middleware.ts
```diff
@@ -0,0 +1,438 @@
...
+        if (!accessToken) {
+            throw new Error("No access token found in the request");
+        }
+        const signOutURL = new URL(
+            `${AppInfo.apiBasePath.getAsStringDangerous()}/signout`,
+            AppInfo.apiDomain.getAsStringDangerous()
+        );
+        const refreshRequestHeaders = new Headers();
+        refreshRequestHeaders.append("Content-Type", "application/json");
+        refreshRequestHeaders.append("Cookie", `${REFRESH_TOKEN_COOKIE_NAME}=${accessToken}`);
```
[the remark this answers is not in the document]
bcbogdan: Resolved

lib/ts/nextjs/ssr.ts
```diff
@@ -105,16 +104,25 @@ export default class SuperTokensNextjsSSRAPIWrapper {
      * @returns An object that includes session context value and the status of the session ('valid' | 'expired' | 'invalid')
      * If the status is 'invalid' or 'expired' then the users should be considered as unauthenticated
      **/
-    static async getServerActionSession(
-        cookies: CookiesStore
+    static async getServerActionSessionWithoutClaims(
+        cookies: CookiesStore,
+        requireAuth: boolean = false
```
porcellus:
```suggestion
        requireAuth: boolean = true
```

lib/ts/nextjs/ssr.ts
```diff
@@ -105,16 +104,25 @@ export default class SuperTokensNextjsSSRAPIWrapper {
...
-            return { session: session as LoadedSessionContext, status: "valid" };
-        } else if (["tokens-do-not-match", "front-token-expired", "access-token-not-found"].includes(state)) {
+            return { session: session as SSRSessionContext, status: "valid" };
+        }
+
+        if (requireAuth) {
+            const redirectPath = cookies.get(CURRENT_PATH_COOKIE_NAME)?.value || "/";
+            const authPagePath = getAuthPagePath(redirectPath);
+            return redirect(authPagePath);
+        }
```
porcellus: Does this work here?

PR #899, opened 2025-05-26
feat: Add init method support, better support for dynamic route handlers and added support for exporting logic from plugins

coolbueb:
## Summary of change

(A few sentences about this PR)

## Related issues

-   Link to issue1 here
-   Link to issue1 here

## Test Plan

## Documentation changes

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

-   [ ] Item1
-   [ ] Item2

### DIFF
--- a/lib/build/genericComponentOverrideContext.js
+++ b/lib/build/genericComponentOverrideContext.js
@@ -1342,7 +1342,10 @@ var getTranslationFunction = function () {
     getCurrentLanguageFromCookie().then(function (cookieLanguage) {
         if (cookieLanguage) setLanguage(cookieLanguage);
     });
-    var translate = function (key) {
+    var translate = function (key, replacements) {
+        if (replacements === void 0) {
+            replacements = {};
+        }
         if (userTranslationFunc) {
             return userTranslationFunc(key);
         }
@@ -1355,7 +1358,9 @@ var getTranslationFunction = function () {
                 }
                 return key;
             }
-            return res;
+            return res.replace(/{(\w+)}/g, function (match, p1) {
+                return replacements[p1] || match;
+            });
         }
         throw new Error("Should never come here");
     };
--- a/lib/build/translation/translationHelpers.d.ts
+++ b/lib/build/translation/translationHelpers.d.ts
@@ -23,4 +23,6 @@ export declare class TranslationController implements TranslationControlEventSou
 }
 export declare function saveCurrentLanguage(language: string, cookieDomain: string | undefined): Promise<void>;
 export declare function getCurrentLanguageFromCookie(): Promise<string | null>;
-export declare const getTranslationFunction: <T extends string>(...stores: TranslationStore[]) => (key: T) => string;
+export declare const getTranslationFunction: <T extends string>(
+    ...stores: TranslationStore[]
+) => (key: T, replacements?: Record<string, string>) => string;
--- a/lib/ts/translation/translationHelpers.ts
+++ b/lib/ts/translation/translationHelpers.ts
@@ -73,7 +73,9 @@ export async function getCurrentLanguageFromCookie(): Promise<string | null> {
     }
 }
 
-export const getTranslationFunction = <T extends string>(...stores: TranslationStore[]): ((key: T) => string) => {
+export const getTranslationFunction = <T extends string>(
+    ...stores: TranslationStore[]
+): ((key: T, replacements?: Record<string, string>) => string) => {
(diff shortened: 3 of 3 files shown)


lib/ts/types.ts
```diff
@@ -538,8 +543,20 @@ export type SuperTokensPlugin = {
         };
     };
     generalAuthRecipeComponentOverrides?: AuthRecipeComponentOverrideMap;
-    routeHandlers?: PluginRouteHandler[];
+    routeHandlers?:
+        | ((
+              config: Omit<SuperTokensConfig, "experimental" | "recipeList">,
+              allPlugins: Pick<SuperTokensPlugin, "id" | "version" | "exports">[],
```
porcellus, line 549: We should export this to a reusable type or re-use SuperTokensPublicPlugin

lib/ts/types.ts
```diff
@@ -528,8 +528,13 @@ export type SuperTokensPlugin = {
     version?: string;
     compatibleAuthReactSDKVersions?: string | string[]; // match the syntax of the engines field in package.json
     compatibleWebJSSDKVersions?: string | string[]; // match the syntax of the engines field in package.json
+    init?: (
+        config: Omit<SuperTokensConfig, "experimental" | "recipeList">,
+        allPlugins: Pick<SuperTokensPlugin, "id" | "version" | "exports">[],
+        sdkVersion: string
+    ) => Promise<void> | void;
```
porcellus, line 535: I don't think there's a usecase for an async init function, so we should just drop the promise part for now.

porcellus (changes requested):

lib/ts/utils.ts
```diff
@@ -553,3 +559,53 @@ export function useRethrowInRender() {
...
+            functions: (oI: any, builder: OverrideableBuilder<any>) => {
+                for (const layer of functionLayers) {
+                    builder.override(layer as any);
+                }
+                return oI as any;
+            },
+        };
+    }
+
+    return _config;
```
porcellus, line 594: This doesn't do anything in auth-react.

lib/ts/recipe/totp/recipe.tsx
```diff
@@ -75,36 +75,43 @@ export default class TOTP extends RecipeModule<
...
+                overrideMaps
             ): RecipeModule<
                 GetRedirectionURLContext,
                 PreAndPostAPIHookAction,
                 OnHandleEventContext,
                 NormalisedConfig
             > => {
+                const normalisedConfig = normaliseMultiFactorAuthFeature(
+                    applyPlugins(TOTP.RECIPE_ID, config, overrideMaps ?? [])
+                );
```
porcellus, line 92: We do not need to do this I think (see my other comment on `applyPlugins`

porcellus (changes requested):

lib/ts/recipe/session/index.ts
```diff
@@ -110,6 +110,11 @@ export default class SessionAPIWrapper {
     }
 
     static ComponentsOverrideProvider = RecipeComponentsOverrideContextProvider;
+
+    // todo add exports like this to all recipes
```
porcellus, line 114: TODO

lib/ts/recipe/passwordless/recipe.tsx
```diff
@@ -103,36 +102,39 @@ export default class Passwordless extends AuthRecipe<
...
             authReact: (
-                appInfo: NormalisedAppInfo
+                appInfo
             ): RecipeModule<
                 GetRedirectionURLContext,
                 PreAndPostAPIHookAction,
                 OnHandleEventContext,
                 NormalisedConfig
             > => {
+                const normalisedConfig = normalisePasswordlessConfig(config);
```
porcellus, line 115: These recipe changes can be reverted entirely; they do not actually change anything but make us call the normalisation function twice.

lib/ts/superTokens.tsx
```diff
@@ -201,7 +235,12 @@ export default class SuperTokens {
 
         for (const plugin of finalPluginList) {
             if (plugin.config) {
-                config = { ...config, ...plugin.config(config) };
+                // prevent override of appInfo
+                // doing it like this because we don't want to override the appInfo and we can't make sure the plugin won't return it
+                // @ts-ignore
+                const { appInfo, ...pluginConfig } = plugin.config(getPublicConfig(config)) || {};
```
porcellus, line 241: I don't think this does what the comment says.

porcellus (changes requested):
I think we are also (still) missing a way to trigger the success redirection from plugins.

PR #923, opened 2026-02-12
fix: Include the rid header in ssr refresh requests

bcbogdan:
## Summary of change

Updates the ssr session refresh behavior to include the rid header in the request. This way the backend sdk will not through csrf related errors during refresh attempts.

## Related issues

-   Link to issue1 here
-   Link to issue1 here

## Test Plan

## Documentation changes

    -   In `package.json`
    -   In `package-lock.json`
    -   In `lib/ts/version.ts`

-   [ ] Item1
-   [ ] Item2

### DIFF
--- a/lib/ts/nextjs/middleware.ts
+++ b/lib/ts/nextjs/middleware.ts
@@ -180,10 +180,15 @@ async function revokeSession(request: Request): Promise<Response | void> {
             `${AppInfo.apiBasePath.getAsStringDangerous()}/signout`,
             AppInfo.apiDomain.getAsStringDangerous()
         );
+
         const signoutRequestHeaders = new Headers();
         signoutRequestHeaders.append("Content-Type", "application/json");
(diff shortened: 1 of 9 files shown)


lib/ts/nextjs/middleware.ts
```diff
@@ -322,6 +327,10 @@ async function fetchNewTokens(request: Request): Promise<{
     const headerRefreshToken = getCookie(request, REFRESH_TOKEN_HEADER_SESSION_COOKIE_NAME);
     const refreshRequestHeaders = new Headers();
     refreshRequestHeaders.append("Content-Type", "application/json");
+    const rid = request.headers.get("rid");
+    if (rid) {
+        refreshRequestHeaders.append("rid", rid);
+    }
```
porcellus: I think these (both changes) should always have the `rid` of `session`. We may want to check the input, but passing it on doesn't necessarily make sense.

porcellus (changes requested):
Please update the version number (setting the release date in the changelog as well)

porcellus (approved):

